PR #1060, opened 2026-05-07
refactor: streamline locale mapping process in createContentfulMapper…

yashin4112:
This PR improves the performance of locale storage by reducing the time required to save locale data in the database.


umesh-more-cstk:
@yashin4112 can you please tell ,we what we did here

vikrantraut-cstk (approved):

yashin4112:
> @yashin4112 can you please tell ,we what we did here

I reduced the time of saving lcoale in db before the call was after mapper call so once mapper generation done we give permission to user to go on 2nd step but there might be situation lcoale didnt save and thats why we just get no options in dropdown for locale mapping in 2nd step so i just moved the call to before mapper generation so we will not get no option in locale mapping dropdown.

umesh-more-cstk (approved):

PR #1064, opened 2026-05-08
Fix empty locale dropdown on step 2 for Drupal

yashin4112:
Fixes empty locale dropdown on step 2 by moving the locale save API call before mapper creation. This change applies only to Drupal.

Ticket: https://contentstack.atlassian.net/browse/CMG-913

### DIFF
--- a/upload-api/src/services/drupal/index.ts
+++ b/upload-api/src/services/drupal/index.ts
@@ -24,6 +24,33 @@ const createDrupalMapper = async (
 
     const localeData = await extractLocale(config);
 
+     // 🔧 CRITICAL: Always normalize to lowercase before saving
+     const localeArray = Array.from(localeData).map((locale: any) => {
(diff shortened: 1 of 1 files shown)


upload-api/src/services/drupal/index.ts
```diff
@@ -24,6 +24,33 @@ const createDrupalMapper = async (
...
+    const mapRes = await axios.request(mapperConfig);
+    if (mapRes?.status == 200) {
+      logger.info('Legacy CMS', {
+        status: HTTP_CODES?.OK,
+        message: HTTP_TEXTS?.LOCALE_SAVED
+      });
+    }
+
     // Extract taxonomy vocabularies and save to drupalMigrationData
     await extractTaxonomy(config.mysql);
```
vikrantraut-cstk: config?.

yashin4112: resolved

umesh-more-cstk (approved):

shradha-nahar (approved):

PR #1108, opened 2026-07-03
CMG-1057, CMG-1058 | Language Mapper and Locale Migration Fixes

chetan-contentstack:
## 🔗 Jira Ticket

[CMG-1057](https://contentstack.atlassian.net/browse/CMG-1057) · [CMG-1058](https://contentstack.atlassian.net/browse/CMG-1058)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 🧹 Chore / Dependency Update
- [ ] 📝 Documentation

---

### What changed?

**CMG-1057 — Language mapper (Step 2) fixes for delta restart:**
- Already-mapped locales no longer appear as selectable options in dropdowns on restart iteration
- Add Language button no longer gets stuck disabled after adding one language — now stays available until all source locales are mapped
- Newly-added rows in a restart iteration remain editable until the user advances past Step 2, so a wrong selection can be corrected without being locked immediately

**CMG-1058 — Delta migration correctness and UI fixes:**
- New locales added in a delta iteration now correctly **localize** existing Contentstack entries rather than creating duplicate new entries
- Entry mapper Save button is now visible whenever there is table data, not only when total counts are non-zero
- Entry mapper footer is sticky so the Save button is not clipped below the viewport
- Step 1 → Step 3 skip guard tightened to also require all source locales to be mapped, preventing Step 2 (locale mapping) from being bypassed

### Why?

On a delta restart with extra locales added, the language mapper was showing already-consumed locales in dropdowns and blocking the user from adding more than one new language. On migration, new-locale entries were being imported as fresh entries (new UIDs) instead of being added as localizations to the entries created in the first iteration, breaking the localization relationship entirely.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [x] `ui` — React frontend
- [ ] `upload-api` — Upload API server
- [ ] `docker` / `docker-compose`
- [ ] CI / GitHub Actions workflows
- [ ] Environment variables / config

---

## 🧪 How to Test

**CMG-1057:**
1. Complete a first iteration with only the master locale mapped
2. Start a second iteration (restart) on Step 2
3. Verify already-mapped locales do not appear in source/destination dropdowns
4. Add two new languages — verify the Add Language button stays enabled after the first addition
5. Select both dropdowns for a newly added row — verify the row stays editable (not immediately disabled) so you can correct a wrong pick

**CMG-1058:**
[...]

PR #1117, opened 2026-07-21
CMG-1057 | Restart Locale Mapper Add Language dropdown

chetan-contentstack:
## 🔗 Jira Ticket

[CMG-1057](https://contentstack.atlassian.net/browse/CMG-1057)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- **Add Language button no longer locks after a single addition on delta restart.** `filledMappingCount` had a stale `sourceLocaleLabels.has(k)` filter meant to defend against `handleSelectedSourceLocale`'s old fallback-key path (source-first pick writing under `source.label.toLowerCase()`). That path has been gone for a while — source-first pick now no-ops in `selectedMappings` until CS is also picked — but the filter kept firing for **legitimate** CS locale keys whose codes match a source locale code (e.g. CS `fr` mapped to source `fr`, or `ja` → `ja`), dropping them from the count and freezing the button at 1.
- **Add Language stays enabled while any source is still unmapped**, regardless of whether earlier rows are fully filled. The old `hasIncompleteRow` gate forced a strict "fill the current row before adding another", which meant the button toggled off after every Add and only came back when the row was fully mapped. Users can now spawn multiple empty rows and fill them in any order.
- **Row count is capped at `sourceLocales.length`.** Prevents hammering Add and spawning empty rows that can never be filled — for 3 sources you get at most 3 rows on screen, whether they're mapped yet or not. Deleting a row frees up a slot again.

### Why?

On delta-restart with sources like `['en', 'fr', 'ja']`, a user could map the first additional locale, and Add Language would then stay disabled forever — trapping the user with 1 unmapped source they couldn't reach. Root cause was the stale filter above. While there, the "fully-fill-before-next-Add" behaviour was tightening the UX for no benefit; loosening it and adding a hard row cap gave a cleaner mapping flow.

---

## 🧩 Affected Areas

- [ ] `api` — Node.js backend
- [x] `ui` — React frontend
- [ ] Other:

Single file: [ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx](ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx) (Add Language `disabled` predicate only).

---

## 🧪 How to Test

Prerequisite: a delta-migration project that finished iteration 1 with ≥ 3 source locales (e.g. Sitecore/Contentful project with source `['en', 'fr', 'ja']` and only `en` mapped as master).

1. Restart migration on the project so it enters iteration 2, then land on **Step 2 · Configure Destination Stack**.
2. Confirm the master row (`en-us (master) ↔ en`) renders with its saved source populated, and **Add Language is enabled**.
3. Click **Add Language** → new empty row appears. Verify Add Language **stays enabled** (does not disable while the row is empty).
4. Pick a CS locale whose code matches a source (e.g. CS `fr` → source `fr`). Add Language still enabled.
5. Click **Add Language** again for the last unmapped source. Two empty rows now on screen. Row cap = 3 (master + 2), so Add Language should now be **disabled**.
6. Fill in the last row (CS `ja` → source `ja`). All sources mapped → button remains disabled (correct — no more sources to add).
7. Delete one non-master row: Add Language re-enables. Restore it: disables again.

**Expected result:**
- Master row auto-populates on restart.
- Add Language is enabled whenever at least one source is unmapped AND row count < total sources.
- Row count never exceeds `sourceLocales.length`.
- No CS-locale/source-code collision (`fr` ↔ `fr`, `ja` ↔ `ja`) locks the button.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
| Add Language disables permanently after the first addition when CS locale code matches a source code; 1 source stays unmappable. | Add Language enabled while unmapped sources remain; row count capped at `sourceLocales.length`; users can spawn multiple empty rows and fill in any order. |

---

- Follow-up to #1108 (CMG-1057/1058 initial fix) — the earlier fix introduced the `sourceLocaleLabels` filter as defence for a code path that has since been removed; this PR retires the filter and simplifies the gate.

---

## ✅ Author Checklist

- [x] `.env` / `example.env` updated if new environment variables were added *(n/a)*
- [ ] Existing tests pass locally (`npm test`) *(run before requesting review)*
- [x] New tests written (or not applicable — explain why) — **n/a: pure UI-gate predicate change, covered by the manual test plan above; no pure logic extracted to a utility.**
- [x] `README.md` / docs updated if behaviour changed *(n/a — internal UI gating only)*
- [ ] Talisman pre-push scan passes (no secrets flagged) *(will confirm on push)*

---

## 👀 Reviewer Notes

- The whole change is contained in a single IIFE inside the `Button`'s `disabled={...}` prop — the non-restart branch above it is untouched.
- Please sanity-check the `rowCapReached` gate: it uses `sourceLocales?.length` (hydrated state), same source of truth as the existing `allSourcesMapped` check, so both bounds move together. If Redux `sourceLocale` hasn't hydrated yet, `sourcesNotReady` already disables the button — the cap doesn't need its own guard.
- The retired `sourceLocaleLabels` filter's only caller was `filledMappingCount`, and `filledMappingCount` itself is gone with `hasIncompleteRow`, so no dead code is left behind.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -67,7 +67,8 @@
     "lodash": "^4.18.1",
     "dompurify": ">=3.4.11",
     "form-data": ">=4.0.6",
-    "ws": ">=8.21.0"
+    "ws": ">=8.21.0",
+    "esbuild": "^0.28.1"
   },
   "eslintConfig": {
     "extends": [
--- a/ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx
+++ b/ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx
@@ -887,40 +887,20 @@ const LanguageMapper = ({stack, uid} :{ stack : IDropDown, uid : string}) => {
                   newMigrationData?.project_current_step > 2
                 );
               }
-              // Restart iteration: button must remain available so users can add a new
-              // destination locale before the delta run. Block when (a) there's already an
-              // in-progress row not yet backed by a completed mapping (avoid stacking empties)
-              // OR (b) every available source locale is already mapped — otherwise clicking
-              // Add Language would surface a row whose source dropdown has no unmapped option
-              // to pick (e.g. a single-locale source where `en` is already in use).
+              // Restart iteration: keep Add Language enabled as long as there is still an
+              // un-mapped source locale, regardless of whether prior rows are fully filled.
+              // Users can add multiple empty rows and fill them in any order.
               //
-              // Row-completeness derives from Redux (`localeMapping`) rather than
-              // `cmsLocaleOptions[i].value`, because the row-object's `value` field is only
-              // populated on rebuild — the per-row handlers (handleSelectedCsLocale /
-              // handleSelectedSourceLocale) update `selectedMappings` → Redux but never write
-              // back into `cmsLocaleOptions`, so a freshly-filled row would otherwise still
-              // read as "empty" here and lock Add Language forever after one add.
-              //
-              // Treat the master row as always-filled: its value is auto-managed by the
-              // parent (stack master locale), and on a delta-restart race the source can be
-              // momentarily blank in Redux before the sync-from-Redux effect merges it back.
-              // Counting master as filled prevents that transient state from disabling the
-              // button on the very first render after restart.
+              // Block when:
+              //   (a) `sourceLocales` hasn't hydrated from Redux yet — new rows would have
+              //       nothing to pick in the source dropdown,
+              //   (b) every available source is already mapped — same reason,
+              //   (c) the number of rows on screen already equals the number of source
+              //       locales — otherwise a user hammering Add would spawn empty rows that
+              //       can never be filled (only N sources exist for N rows). The row cap
+              //       kicks in even before any mapping happens (e.g. 3 sources → max 3 rows,
+              //       whether they're mapped yet or not).
               const savedMapping = newMigrationData?.destination_stack?.localeMapping || {};
-              // Exclude keys that are source locale codes — handleSelectedSourceLocale
-              // previously wrote under a fallback key (source label lowercased) when the
-              // user picked source before CS, inflating the count and re-enabling Add Language
-              // for an incomplete row. Filter those out so only proper CS locale keys count.
-              const sourceLocaleLabels = new Set(
-                sourceLocales?.map((l: { label: string }) => l.label) ?? []
-              );
-              const filledMappingCount = Object.entries(savedMapping).filter(([k, v]) => {
-                const isMasterKey = typeof k === 'string' && k.endsWith('-master_locale');
-                if (isMasterKey) return true;
-                if (sourceLocaleLabels.has(k)) return false;
-                return typeof v === 'string' && v.length > 0;
-              }).length;
-              const hasIncompleteRow = (cmsLocaleOptions?.length ?? 0) > filledMappingCount;
               const mappedSources = new Set(
                 Object.values(savedMapping).filter(
                   (v): v is string => typeof v === 'string' && v?.length > 0
@@ -933,7 +913,9 @@ const LanguageMapper = ({stack, uid} :{ stack : IDropDown, uid : string}) => {
               const totalSources = sourceLocales?.length ?? 0;
               const sourcesNotReady = totalSources === 0;
               const allSourcesMapped = totalSources > 0 && mappedSources.size >= totalSources;
-              return hasIncompleteRow || sourcesNotReady || allSourcesMapped;
+              const rowCapReached =
+                totalSources > 0 && (cmsLocaleOptions?.length ?? 0) >= totalSources;
+              return sourcesNotReady || allSourcesMapped || rowCapReached;
             })()}
           >
             Add Language
--- a/upload-api/package.json
+++ b/upload-api/package.json
@@ -35,7 +35,7 @@
     "@types/wordpress__blocks": "^12.5.18",
     "@types/xml2js": "^0.4.14",
     "@typescript-eslint/eslint-plugin": "^8.56.1",
-    "@vitest/coverage-v8": "^4.0.18",
+    "@vitest/coverage-v8": "^4.1.10",
     "eslint": "^9.0.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.1.3",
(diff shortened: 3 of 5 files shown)


vikrantraut-cstk (approved):

shradha-nahar (approved):

ui/package.json
```diff
@@ -67,7 +67,8 @@
     "lodash": "^4.18.1",
     "dompurify": ">=3.4.11",
     "form-data": ">=4.0.6",
-    "ws": ">=8.21.0"
+    "ws": ">=8.21.0",
+    "esbuild": "^0.28.1"
```
umesh-more-cstk, line 71: **blocker (scope):** This `esbuild` override is unrelated to CMG-1057 (a locale-mapper UI gate). It's correctly placed in `overrides` (not runtime deps), but it silently moves the build toolchain 0.27.3 → 0.28.1 inside a bugfix PR and adds ~26 platform packages of lockfile churn.

The description says *"Single file"* / *"Affected Areas: ui only"*.

**Suggestion:** split dependency bumps into a separate chore PR, or if a security pin must ride along, call it out in the description and check the "Dependency Update" box.

upload-api/package.json
```diff
@@ -35,15 +35,15 @@
...
+    "@vitest/coverage-v8": "^4.1.10",
     "eslint": "^9.0.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.1.3",
     "nodemon": "^3.1.10",
     "rimraf": "^6.0.1",
     "ts-node": "^10.9.2",
     "typescript": "^5.3.3",
-    "vitest": "^4.0.18"
+    "vitest": "^4.1.10"
```
umesh-more-cstk, line 46: **blocker (scope):** `vitest` (and `@vitest/coverage-v8` above) are bumped to `^4.1.10` here in **upload-api** — but upload-api isn't checked in the PR's Affected Areas, and this is unrelated to CMG-1057. Pulls in oxc/rolldown lockfile churn too.

**Suggestion:** move these into a separate dependency PR so the bugfix stays isolated and cleanly revertable.

ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx
```diff
@@ -887,40 +887,20 @@ const LanguageMapper = ({stack, uid} :{ stack : IDropDown, uid : string}) => {
...
-              const filledMappingCount = Object.entries(savedMapping).filter(([k, v]) => {
-                const isMasterKey = typeof k === 'string' && k.endsWith('-master_locale');
-                if (isMasterKey) return true;
-                if (sourceLocaleLabels.has(k)) return false;
-                return typeof v === 'string' && v.length > 0;
-              }).length;
-              const hasIncompleteRow = (cmsLocaleOptions?.length ?? 0) > filledMappingCount;
               const mappedSources = new Set(
                 Object.values(savedMapping).filter(
                   (v): v is string => typeof v === 'string' && v?.length > 0
```
umesh-more-cstk, line 906: **nit:** `v` is already narrowed to `string` by `typeof v === 'string'`, so `v?.length` can be `v.length`. Harmless, but the `?.` reads as if `v` could still be nullish here.

ui/src/components/DestinationStack/Actions/LoadLanguageMapper.tsx
```diff
@@ -933,7 +913,9 @@ const LanguageMapper = ({stack, uid} :{ stack : IDropDown, uid : string}) => {
               const totalSources = sourceLocales?.length ?? 0;
               const sourcesNotReady = totalSources === 0;
               const allSourcesMapped = totalSources > 0 && mappedSources.size >= totalSources;
-              return hasIncompleteRow || sourcesNotReady || allSourcesMapped;
+              const rowCapReached =
+                totalSources > 0 && (cmsLocaleOptions?.length ?? 0) >= totalSources;
+              return sourcesNotReady || allSourcesMapped || rowCapReached;
```
umesh-more-cstk, line 918: **question (medium):** Both `rowCapReached` and `allSourcesMapped` assume every value in `savedMapping` corresponds to a source still present in `sourceLocales`. If a delta restart can carry a **stale mapping for a source that no longer exists** (source set shrank between iterations), this re-locks the button — the exact failure class this PR fixes, via a new vector:

```
Iter 1 mapped: { en-master_locale: en, de: de }   // 'de' later removed from source export
Iter 2 sourceLocales: [en, fr, ja]  → totalSources = 3
→ rebuilt = 2 rows (master, de);  mappedSources = {en, de}
→ user Adds a row (cap 3 reached), maps it to fr
→ mappedSources = {en, de, fr} → size 3 ≥ 3 → allSourcesMapped = true → disabled
→ 'ja' is now unreachable
```

Is `localeMapping` guaranteed to be reconciled against the current `sourceLocales` on restart? If a removed source can linger, harden by intersecting with the live source set:

```ts
const sourceSet = new Set(sourceLocales?.map((l) => l.value) ?? []);
const mappedSources = new Set(
  Object.values(savedMapping).filter(
    (v): v is string => typeof v === 'string' && v.length > 0 && sourceSet.has(v)
  )
);
```

If reconciliation is guaranteed upstream, no change needed — just confirm it.

umesh-more-cstk (commented):
## Review — CMG-1057 Add Language dropdown

The isolated `disabled`-predicate change is a clean, correct fix. Retiring the `sourceLocaleLabels`/`filledMappingCount` filter is the right call — it genuinely mis-fired on legitimate CS keys whose code matched a source code (`fr`↔`fr`, `ja`↔`ja`). Cap-consistency holds: `rowCapReached` and `allSourcesMapped` both derive from `sourceLocales`, and the delete path clears `selectedMappings`→Redux so the button re-enables after a row delete.

**Verdict: Comment** (the code change alone I'd approve; my one blocking-class reservation is scope — see the dependency comments).

### Blocking
- **Unrelated dependency/lockfile churn vs. stated scope** — the PR says *"single file / ui only"* but also bumps `esbuild` (ui overrides) and `vitest`/`@vitest/coverage-v8` in **upload-api** (not even checked in Affected Areas), with large lockfile churn in both packages. See inline comments on `ui/package.json` and `upload-api/package.json`.

### Non-blocking
- rowCap/allSourcesMapped drift edge case (inline, `LoadLanguageMapper.tsx`)
- redundant optional chaining (inline)
- **Stale comment (nit):** `LoadLanguageMapper.tsx:419` still references `filledMappingCount`, which this PR deleted. The `selectedLocaleKey` guard it documents is still valid — just retarget the wording (e.g. "…would create an orphan mapping key and mis-count mapped sources"). Couldn't anchor inline since that line is outside the diff.

PR #1118, opened 2026-07-21
fix: improve responsive behavior of AssetMapper and EntryMapper table…

yashin4112:
## 🔗 Jira Ticket

[CMG-1044](https://contentstack.atlassian.net/browse/CMG-1044)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- Made the **Entry Mapper** and **Asset Mapper** tables (Map Entry step) fully responsive so the pagination bar and Save footer stay visible and reachable at any browser zoom level.
- Reworked the height chain into a bounded flex layout: `.entry-asset-mapper` is now pinned to the real step-content area (`calc(100vh - 246px)`, derived from the stepper's own `.stepContent` height minus its fixed header padding), so the subtree is self-bounding instead of growing to content height.
- The table body height is now **measured** off the bounded outer box (minus the toggle, search row, pagination bar and Save footer) and fed into `InfiniteScrollTable`'s `tableHeight` prop via a `ResizeObserver`, so react-window's virtual viewport matches the visible area and its native scroll works. Anchored on the outer box (never on `.Table`/`.Table__body`) to avoid a measurement feedback loop.
- Replaced the fixed `calc(100vh - Npx)` / hardcoded `tableHeight={400}` sizing with flex + measured height; the table body scrolls internally while pagination + Save stay pinned below.
- Added the `entry-asset-mapper` class to the Entry/Asset toggle wrapper (it had none, so its flex layout rules never applied) and removed dead `.asset-mapper-container .Table` fixed-height rules.

### Why?

On zoom (110%+), the fixed-pixel/`100vh`-based table height didn't shrink with the effective viewport, so the pagination bar slid below the Save footer and the Save button became unreachable — the UI locked with no way to scroll to it. This makes the layout distribute height correctly and keep the controls in view at 100%, 110% and beyond.

---

## 🧩 Affected Areas

- [ ] `api` — Node.js backend
- [x] `ui` — React frontend
- [ ] Other:

---

## 🧪 How to Test

1. Open a delta-iteration project and go to the **Map Entry** step (Entries / Assets tabs).
2. Select a content type with enough entries to paginate.
3. Set browser zoom to **100%**, then **110%** and higher.

**Expected result:** At every zoom level the table body scrolls internally, and the pagination bar and **Save** button stay visible and reachable at the bottom (no overflow, no locked UI). Repeat on the **Assets** tab.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
|        |       |

---

-

---

## ✅ Author Checklist

- [ ] `.env` / `example.env` updated if new environment variables were added — N/A (no env changes)
- [x] Existing tests pass locally (`npm test`)
- [ ] New tests written (or not applicable — explain why) — N/A, CSS/layout-only change with no unit-testable logic
- [ ] `README.md` / docs updated if behaviour changed — N/A
- [x] Talisman pre-push scan passes (no secrets flagged)

---

## 👀 Reviewer Notes

The table body height is driven by a measured JS value (`tableHeight` prop), **not** CSS, because `InfiniteScrollTable`/react-window sizes its virtual scroll viewport from that number. The measurement deliberately anchors on the bounded outer box (`.entry-asset-mapper`) and never reads `.Table`/`.Table__body` — reading those creates a runaway feedback loop (we saw `tableHeight` blow up to ~9966px during development). Please sanity-check the `calc(100vh - 246px)` reserve if the surrounding stepper chrome changes.

---

### DIFF
--- a/ui/src/components/ContentMapper/entryMapper.tsx
+++ b/ui/src/components/ContentMapper/entryMapper.tsx
@@ -477,13 +477,14 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
       ),
       accessor: accessorCall,
       id: 'uuid',
-      width: '250px',
+      width: '340px',
     },
     {
       disableSortBy: true,
       Header: <span>{`${newMigrationData?.legacy_cms?.selectedCms?.title} UIDs:`}</span>,
       accessor: accessorForCMSUid,
-      id: '1'
+      id: '1',
+      width: '360px',
     },
     {
       disableSortBy: true,
@@ -493,11 +494,41 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
     }
   ];
 
-  // Must match the .Table__body height in index.scss so the react-window list is exactly
-  // as tall as the scroll body. Leave ~140px below for the search row, pagination bar and
-  // Save footer; the body scrolls internally so all rows of a page stay reachable.
-  const calcHeight = () => window.innerHeight - 520;
-  const tableHeight = calcHeight();
+  // react-window sizes its virtual scroll viewport from this JS number, not CSS. The layout
+  // flexes the table body to fill the available space (responsive on zoom), so we measure
+  // that rendered body height and feed it back here — otherwise react-window renders a
+  // fixed-height viewport that doesn't match the flexed body and scrolling breaks.
+  const [tableHeight, setTableHeight] = useState<number>(() => window.innerHeight - 520);
+  useEffect(() => {
+    const measure = () => {
+      // Anchor on the OUTER bounded box (.entry-asset-mapper is capped at calc(100vh-246px)),
+      // NOT on .Table/.Table__body — those are sized BY react-window from this very number, so
+      // reading them creates a runaway feedback loop. Subtract only the fixed chrome that sits
+      // inside the box: the toggle row, the search/panel row, pagination bar and Save footer.
+      const box = document.querySelector('.entry-asset-mapper') as HTMLElement | null;
+      const toggle = document.querySelector('.mapper-view-toggle') as HTMLElement | null;
+      const panel = document.querySelector('.entry-mapper-container .TablePanel') as HTMLElement | null;
+      const footer = document.querySelector('.entry-mapper-table .mapper-footer') as HTMLElement | null;
+      const boxH = box?.clientHeight ?? window.innerHeight - 246;
+      const RESERVE =
+        (toggle?.offsetHeight ?? 0) +
+        (panel?.offsetHeight ?? 64) +
+        (footer?.offsetHeight ?? 65) +
+        56; // pagination bar (fixed) + small buffer
+      const avail = boxH - RESERVE;
+      if (avail > 80) setTableHeight(Math.floor(avail));
+    };
+    measure();
+    const box = document.querySelector('.entry-asset-mapper') as HTMLElement | null;
+    const ro = new ResizeObserver(measure);
+    if (box) ro.observe(box);
+    window.addEventListener('resize', measure);
+    return () => {
+      ro.disconnect();
+      window.removeEventListener('resize', measure);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [contentTypeUid, tableData?.length]);
 
   return (
     isLoading || newMigrationData?.isprojectMapped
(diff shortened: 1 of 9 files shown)


vikrantraut-cstk (approved):

vikrantraut-cstk (dismissed):

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -493,11 +494,41 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+    const box = document.querySelector('.entry-asset-mapper') as HTMLElement | null;
+    const ro = new ResizeObserver(measure);
+    if (box) ro.observe(box);
+    window.addEventListener('resize', measure);
+    return () => {
+      ro.disconnect();
+      window.removeEventListener('resize', measure);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [contentTypeUid, tableData?.length]);
```
umesh-more-cstk: **cleanup (medium):** This ~30-line measurement effect is duplicated almost verbatim in `assetMapper.tsx` (lines 76–102) — only the `.TablePanel`/`.mapper-footer` selectors differ. Please extract a shared hook so the reserve math and feedback-loop guard live in one place:

```ts
// useMeasuredTableHeight.ts
export function useMeasuredTableHeight(
  deps: unknown[],
  opts: { panelSelector: string; footerSelector: string }
) {
  const [tableHeight, setTableHeight] = useState(() => window.innerHeight - 520);
  useEffect(() => { /* measure + ResizeObserver + resize listener */ }, deps);
  return tableHeight;
}
```

Right now a tweak to the RESERVE constants or the loop-avoidance logic has to be made in two files and can silently drift.

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -619,82 +650,85 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+                  canSearch={true}
+                  totalCounts={totalCounts ?? 0}
+                  data={[...tableData]}
+                  columns={columns}
+                  uniqueKey={'id'}
+                  isRowSelect={true}
+                  fullRowSelect={true}
+                  fetchTableData={fetchData}
+                  tableHeight={tableHeight}
+                  equalWidthColumns={false}
```
umesh-more-cstk, line 656: **question (medium):** This flips `equalWidthColumns` `true → false` and pairs it with fixed `width: '340px'` / `'360px'` on the first two columns (lines 480, 487), which the SCSS then renders as `flex: 0 0 auto` (can't shrink). That's a hard 700px minimum for the first two columns, with only the last column (`flex: 1 1 auto`) able to absorb slack.

At 110%+ zoom / narrow widths — the exact scenario this PR targets — the fields column (`max-width: 75%`) can fall below 700px, so does this reintroduce an overflow, just horizontal (an internal `.Table` h-scroll) instead of vertical? Please confirm at 100/110/125% on a ~1280px screen. Also: is the column widening actually part of CMG-1044, or incidental tuning that came along with `equalWidthColumns=false`? If incidental, a one-line note in the description would help.

umesh-more-cstk (commented):
## Review — CMG-1044 responsive AssetMapper/EntryMapper tables

Solid, well-documented fix for a genuinely awkward layout problem — the extensive comments explaining the feedback-loop avoidance and the `calc(100vh - 246px)` reserve are appreciated and made this reviewable. The bounded-flex + measured-`tableHeight` approach is sound, and observing the fixed-size outer box (not `.Table`/`.Table__body`) correctly avoids the runaway loop. Only one mapper mounts at a time (ternary in `entryAssetMapper.tsx`), so the global `document.querySelector` reads don't cross-match between the two.

**Verdict: Comment — no blockers.** Two things I'd like addressed before merge (both inline): extract the duplicated measurement effect into a shared hook, and confirm the entry-table column-width change doesn't trade the vertical overflow for a *horizontal* one at the same 110%+ zoom this targets.

### Please address before merge
1. **Duplicated measurement effect** (`entryMapper.tsx` ↔ `assetMapper.tsx`) — inline.
2. **`equalWidthColumns` flip + fixed column widths** — verify no horizontal overflow — inline.

### Non-blocking nits
- `box?.clientHeight ?? …` doesn't catch `0`; `avail > 80` leaves a stale height at extreme zoom (inline).
- Magic numbers `246`/`520`/`64`/`65`/`56` are duplicated across both files and the SCSS. The initial state uses `innerHeight - 520` but the box fallback uses `innerHeight - 246` — worth a shared named constant, and the two seeds reconciled.
- The `window.addEventListener('resize', measure)` is largely redundant with the `ResizeObserver` on a `100vh`-based box (both fire on zoom). Harmless, just noise.
- `index.scss`: the removed `.Table` rule leaves a double blank line, and the file has no trailing newline (lint should catch both).

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -493,11 +494,41 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+  // that rendered body height and feed it back here — otherwise react-window renders a
+  // fixed-height viewport that doesn't match the flexed body and scrolling breaks.
+  const [tableHeight, setTableHeight] = useState<number>(() => window.innerHeight - 520);
+  useEffect(() => {
+    const measure = () => {
+      // Anchor on the OUTER bounded box (.entry-asset-mapper is capped at calc(100vh-246px)),
+      // NOT on .Table/.Table__body — those are sized BY react-window from this very number, so
+      // reading them creates a runaway feedback loop. Subtract only the fixed chrome that sits
+      // inside the box: the toggle row, the search/panel row, pagination bar and Save footer.
+      const box = document.querySelector('.entry-asset-mapper') as HTMLElement | null;
```
umesh-more-cstk: **robustness (low):** Measuring via `document.querySelector` on hard-coded venus class names (`.TablePanel`, `.mapper-footer`, and elsewhere `.Table__body`) means a venus class rename or markup change silently falls back to the magic constants (`64`/`65`) with no error — the layout would quietly regress. Since `entryMapper` already holds `tableWrapperRef`, consider scoping the box/footer reads to refs (or `box.querySelector(...)` off the wrapper ref) so a missing node is at least contained and debuggable.

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -493,11 +494,41 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+      const panel = document.querySelector('.entry-mapper-container .TablePanel') as HTMLElement | null;
+      const footer = document.querySelector('.entry-mapper-table .mapper-footer') as HTMLElement | null;
+      const boxH = box?.clientHeight ?? window.innerHeight - 246;
+      const RESERVE =
+        (toggle?.offsetHeight ?? 0) +
+        (panel?.offsetHeight ?? 64) +
+        (footer?.offsetHeight ?? 65) +
+        56; // pagination bar (fixed) + small buffer
+      const avail = boxH - RESERVE;
+      if (avail > 80) setTableHeight(Math.floor(avail));
```
umesh-more-cstk: **nit (low):** Two small edge cases here:
- `box?.clientHeight ?? (window.innerHeight - 246)` — `??` only catches `null`/`undefined`, not `0`. If the wrapper is momentarily 0-height (measured before layout settles), `boxH` is `0`, not the fallback, so `avail` goes negative. It's caught by `avail > 80` so nothing breaks, but the fallback never actually fires as intended. Use `box?.clientHeight || (window.innerHeight - 246)` if you want `0` to fall back.
- When `avail <= 80` the height simply isn't updated, so `tableHeight` keeps its previous (possibly large initial) value — at extreme zoom that could re-expose the overflow. Consider clamping to a small min (`setTableHeight(Math.max(80, Math.floor(avail)))`) rather than skipping the update.

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: resolved. Created useMeasuredTableHeight.ts

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: resolved.

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: Browser zoom scales px and % by the same factor, so the 340/360px columns and the 75% fields column keep their ratio at every zoom level — zoom alone never triggers the h-scroll. On a 1280px screen the fields column is ~960px, well above the ~768px floor, so no overflow at 100/110/125%. It'd only kick in below ~1024px viewport width, which is outside our supported range.

The fixed widths + equalWidthColumns=false are intentional (part of this PR) — needed so the long UID columns stay readable instead of the equal split squeezing them. Added a note in the PR description.

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: resolved.

umesh-more-cstk (dismissed):

ui/src/components/ContentMapper/useMeasuredTableHeight.ts
```diff
@@ -0,0 +1,94 @@
...
+const BOX_HEIGHT_FALLBACK = 246;
+const BOX_SELECTOR = '.entry-asset-mapper';
+const TOGGLE_SELECTOR = '.mapper-view-toggle';
+
+export function useMeasuredTableHeight(
+  wrapperRef: RefObject<HTMLElement | null>,
+  deps: unknown[],
+  { panelSelector, footerSelector }: MeasuredTableHeightOptions,
+): number {
+  const [tableHeight, setTableHeight] = useState<number>(() => window.innerHeight - 520);
```
umesh-more-cstk: **nit:** The initial `useState` guess has two small robustness gaps before the effect's first `measure()` runs (react-window renders one frame with this value):

1. It's **unclamped** — on a short window `window.innerHeight - 520` is negative and gets handed to react-window's virtual viewport for that first frame. `measure()` clamps to `MIN_USABLE_HEIGHT`, but the initial value doesn't.
2. `520` is inconsistent with the hook's own box model (`BOX_HEIGHT_FALLBACK = 246` + the `reserve`), so the placeholder height is a different magic number than the one the rest of the file reasons about.

Suggest reusing the same floor (and ideally the box fallback) so the pre-measure frame matches the model:
```ts
const [tableHeight, setTableHeight] = useState<number>(
  () => Math.max(MIN_USABLE_HEIGHT, window.innerHeight - 520),
);
```

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -619,82 +621,85 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+                  canSearch={true}
+                  totalCounts={totalCounts ?? 0}
+                  data={[...tableData]}
+                  columns={columns}
+                  uniqueKey={'id'}
+                  isRowSelect={true}
+                  fullRowSelect={true}
+                  fetchTableData={fetchData}
+                  tableHeight={tableHeight}
+                  equalWidthColumns={false}
```
umesh-more-cstk, line 656: **question:** This flips the entry table from `equalWidthColumns={true}` to `false`, paired with fixed widths (`uuid: 340px` L481, CMS UIDs: `360px` L488) and the SCSS switch to `.Table__head__column/.Table__body__column { flex: 0 0 auto }` (index.scss L714). Those columns can no longer shrink, and there's also the row-select column. The fields column is capped at `max-width: 75%` and shares the row with the content-types list column, so on smaller viewports the fixed widths (340 + 360 + select + growing last column) could exceed the available row width and overflow horizontally (`.Table__body` only sets `overflow-y: auto`).

This is a horizontal-layout change bundled into an otherwise vertical-responsiveness fix — can you confirm the columns still fit without a horizontal scrollbar/clip at the minimum supported width (list column visible)?

umesh-more-cstk (commented):
### Review: `fix: improve responsive behavior of AssetMapper and EntryMapper tables` (CMG-1044)

**Verdict: COMMENT (non-gating).** This is a well-engineered layout fix. I traced the full flex chain at the head commit and it holds together: `.entry-asset-mapper` (bounded to `calc(100vh - 246px)`, `overflow:hidden`) → `.step-container` (`flex:1`) → `.table-container` (`flex:1 1 auto`, `align-items/content:stretch`) → `.content-types-fields-wrapper` (flex col) → `.entry-table-wrapper` (`flex:1 1 auto`) → `.entry-mapper-table` → `.Table` + `.mapper-footer`. The measured-height approach (feed react-window's `tableHeight` off the bounded outer box, never off `.Table`/`.Table__body`) is the right way to avoid the feedback loop, and the effect cleanup (ResizeObserver `disconnect` + `resize` listener removal) is correct.

I could not find a blocking issue. One thing I checked and **disproved**: I suspected the `reserve` in `useMeasuredTableHeight` omits the `.locale-select-inline` row and would clip the footer on multi-locale projects — but that row is `position:absolute` (overlaid on `.TablePanel`, which reserves space via `min-height:64px; padding-right:272px`), so it does not consume flex height. No bug there.

**Non-blocking (inline):**
- **nit** — `useMeasuredTableHeight.ts:44`: initial `tableHeight` guess is unclamped and inconsistent with the hook's own box model.
- **question** — `entryMapper.tsx:656`: entry columns switched from equal-width to fixed non-shrinking widths (a horizontal-layout change bundled with the vertical fix) — please confirm no horizontal overflow at the min supported width.

**Out-of-diff heads-ups (not blocking):**
- **Tests / ResizeObserver:** `new ResizeObserver(...)` is called unconditionally, but `ui/tests/setup.ts` does not polyfill it. The only ContentMapper tests today are `*.utils.test.ts` (pure functions) — no test mounts `EntryMapper`/`AssetMapper` — so CI is unaffected now. If a component render test is ever added it will throw `ReferenceError: ResizeObserver is not defined` under jsdom; worth a stub in `tests/setup.ts` preemptively.
- **Description vs diff:** the PR text says it "Added the `entry-asset-mapper` class to the Entry/Asset toggle wrapper (it had none)". In the code, `.entry-asset-mapper` is already on the outer wrapper in `entryAssetMapper.tsx` (`.mapper-view-toggle` is the toggle), and that file is not part of this diff — the description likely reflects an earlier commit on the branch. No code impact.

**Verified good:** dev-only `console.warn` on missing selectors instead of silently falling back to magic constants (directly addresses this repo's "DOM reads by hard-coded class names that silently fall back to magic numbers" concern); DOM reads scoped to `wrapperRef` rather than global `document`; the empty-state height rule (`.Table:has(.EmptyState)`, 4-class specificity) still wins over the new `.entry-mapper-table .Table { height:auto }`, so the empty state is preserved; and the entry/asset `.Table__body` rules don't cross-contaminate. Scope is clean — UI-only, matches the stated Affected Area, no lockfile/dependency churn.

ui/src/components/ContentMapper/useMeasuredTableHeight.ts
yashin4112: resolved.

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: resolved.

umesh-more-cstk (approved):

vikrantraut-cstk (approved):

PR #1120, opened 2026-07-22
fix: always create content type on test migration, disable Save when …

yashin4112:
## 🔗 Jira Ticket

. [CMG-995](https://contentstack.atlassian.net/browse/CMG-995) 
· [CMG-1048](https://contentstack.atlassian.net/browse/CMG-1048)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- **fieldAttacher (`api`):** Added an `isTest` flag. On test migrations the content type is now always (re)created into the test stack, short-circuiting the iteration-based skip logic. Previously the `isTest` block and the iteration block both ran on the same pass, causing each content type to be created twice.
- **EntryMapper (`ui`):** Save button is now disabled when the selected content type has no mappable entry (no row has a Contentstack UID), on top of the existing in-flight-migration lock.

### Why?

- Test migrations must reflect the latest content type schema regardless of whether it already exists — the previous fall-through created content types twice and applied the wrong iteration/skip logic.
- Nothing is savable when a content type has no entry mapped to a Contentstack UID, so the Save action was misleading.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [x] `ui` — React frontend
- [ ] Other:

---

## 🧪 How to Test

1. Run a **test migration** and confirm each content type is created into the test stack exactly once (no double-create), including content types that already exist from a previous iteration.
2. Run a **real migration** and confirm iteration-based skip logic still applies (existing content types are skipped on iteration 2+).
3. In the **Entry Mapper** step, open a content type whose entries all show `-` for Contentstack UID → Save should be disabled.
4. Open a content type with at least one mapped Contentstack UID → Save should be enabled.

**Expected result:** Test migrations create content types once; Save is only enabled when there's something mappable to save.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
|        |       |

---

-

---

## ✅ Author Checklist

- [ ] `.env` / `example.env` updated if new environment variables were added — N/A

---

## 👀 Reviewer Notes

Focus on the branching in `fieldAttacher` — the fix converts the old separate `if (isTest)` + `if (iteration === 1)` blocks into a single `if / else if / else` so test migrations no longer fall through into the iteration logic.

---

### DIFF
--- a/api/src/utils/field-attacher.utils.ts
+++ b/api/src/utils/field-attacher.utils.ts
@@ -6,7 +6,7 @@ import { shouldSkipContentTypeCreation } from "./content-type-checker.utils.js";
 import { sanitizeProjectId, sanitizeStackId } from "./sanitize-path.utils.js";
 import customLogger from "./custom-logger.utils.js";
 
-export const fieldAttacher = async ({ projectId, orgId, destinationStackId, region, user_id, is_sso }: any) => {
+export const fieldAttacher = async ({ projectId, orgId, destinationStackId, region, user_id, is_sso, isTest = false}: any) => {
   const safeProjectId = sanitizeProjectId(projectId);
   if (!safeProjectId) {
     throw new Error("Invalid project identifier");
@@ -45,9 +45,13 @@ export const fieldAttacher = async ({ projectId, orgId, destinationStackId, regi
         })
       }
 
-      if (iteration === 1) {
+      // Test migration: always create the content type into the test stack,
+      // regardless of whether it already exists. The iteration-based skip logic
+      if (isTest) {
         await contenTypeMaker({ contentType, destinationStackId: safeDestinationStackId, projectId: safeProjectId, newStack: projectData?.stackDetails?.isNewStack, keyMapper: projectData?.mapperKeys, region, user_id, is_sso })
-
+      }
+      else if (iteration === 1) {
+        await contenTypeMaker({ contentType, destinationStackId: safeDestinationStackId, projectId: safeProjectId, newStack: projectData?.stackDetails?.isNewStack, keyMapper: projectData?.mapperKeys, region, user_id, is_sso })
       }
       else {
         const shouldSkip = await shouldSkipContentTypeCreation(safeProjectId, contentType?.otherCmsUid, iteration);
--- a/api/src/utils/field-attacher.utils.ts
+++ b/api/src/utils/field-attacher.utils.ts
@@ -45,12 +45,10 @@ export const fieldAttacher = async ({ projectId, orgId, destinationStackId, regi
         })
       }
 
-      // Test migration: always create the content type into the test stack,
-      // regardless of whether it already exists. The iteration-based skip logic
-      if (isTest) {
-        await contenTypeMaker({ contentType, destinationStackId: safeDestinationStackId, projectId: safeProjectId, newStack: projectData?.stackDetails?.isNewStack, keyMapper: projectData?.mapperKeys, region, user_id, is_sso })
-      }
-      else if (iteration === 1) {
+      // Always create the content type on a test migration (skip the iteration-based
+      // dedupe entirely) or on the first real iteration. The skip logic below only
+      // applies to delta iterations of a real migration.
+      if (isTest || iteration === 1) {
         await contenTypeMaker({ contentType, destinationStackId: safeDestinationStackId, projectId: safeProjectId, newStack: projectData?.stackDetails?.isNewStack, keyMapper: projectData?.mapperKeys, region, user_id, is_sso })
       }
       else {
--- a/ui/src/components/ContentMapper/entryMapper.tsx
+++ b/ui/src/components/ContentMapper/entryMapper.tsx
@@ -690,9 +690,12 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
                       // Lock the Save button only while a migration is actively in flight.
                       // Using migrationStarted alone would permanently lock revisits on delta
                       // iterations since migrationStarted stays true after completion.
+                      // Also lock when this content type has no selectable entry (no row is
+                      // mapped to a Contentstack uid), since there's nothing to save.
                       disabled={
-                        !!newMigrationData?.migration_execution?.migrationStarted &&
-                        !newMigrationData?.migration_execution?.migrationCompleted
+                        (!!newMigrationData?.migration_execution?.migrationStarted &&
+                          !newMigrationData?.migration_execution?.migrationCompleted) ||
+                        !tableData?.some((row) => row?._canSelect)
                       }
                       isLoading={isLoadingSaveButton}
                     >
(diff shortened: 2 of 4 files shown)


ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -690,9 +690,12 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
                       // Using migrationStarted alone would permanently lock revisits on delta
                       // iterations since migrationStarted stays true after completion.
+                      // Also lock when this content type has no selectable entry (no row is
+                      // mapped to a Contentstack uid), since there's nothing to save.
                       disabled={
-                        !!newMigrationData?.migration_execution?.migrationStarted &&
-                        !newMigrationData?.migration_execution?.migrationCompleted
+                        (!!newMigrationData?.migration_execution?.migrationStarted &&
+                          !newMigrationData?.migration_execution?.migrationCompleted) ||
+                        !tableData?.some((row) => row?._canSelect)
```
umesh-more-cstk: **blocker:** `tableData` here is only the **current server-paginated page** (default `limit=30`; `fetchEntries` *replaces* `tableData` with each page, and `totalCounts` from `data.count` is the real grand total). So `!tableData?.some((row) => row?._canSelect)` reflects the current page, not the content type.

Concrete failures:
1. A content type with > page-size entries whose mappable (`contentstackEntryUid`-bearing) rows fall on page 2+ shows Save **disabled on open**, contradicting the stated intent (there *is* something mappable to save).
2. After the user checks rows on page 1 (persisted in `rowIds`) and pages to a page with no `_canSelect` rows, Save is disabled even though those pending selections are savable.

The predicate ignores both `totalCounts` and the persisted selection (`rowIds`). Suggested fix — don't derive "has mappable entry" from the current page:
```tsx
disabled={
  (!!newMigrationData?.migration_execution?.migrationStarted &&
    !newMigrationData?.migration_execution?.migrationCompleted) ||
  // savable when the user has a pending selection, or the current page still shows a mappable row
  (Object.keys(rowIds ?? {}).length === 0 &&
    !tableData?.some((row) => row?._canSelect))
}
```
More robust still: have the API return a per-content-type count of mappable entries and gate on that. Either way, please confirm entry lists here can exceed the page size (they can — `rowPerPageOptions` goes to 100 and `totalCounts` is a server grand total).

api/src/utils/field-attacher.utils.ts
```diff
@@ -45,9 +45,13 @@ export const fieldAttacher = async ({ projectId, orgId, destinationStackId, regi
         })
       }
 
-      if (iteration === 1) {
+      // Test migration: always create the content type into the test stack,
+      // regardless of whether it already exists. The iteration-based skip logic
+      if (isTest) {
+        await contenTypeMaker({ contentType, destinationStackId: safeDestinationStackId, projectId: safeProjectId, newStack: projectData?.stackDetails?.isNewStack, keyMapper: projectData?.mapperKeys, region, user_id, is_sso })
```
umesh-more-cstk: **question:** Test migrations now always call `contenTypeMaker` regardless of iteration. `contenTypeMaker` -> `saveContent` **appends** to `content_types/schema.json` via `schemaData.push(ct)` (no dedupe), and `clearStaleEntries` (called just before `fieldAttacher` in the test flow) is scoped to the `entries/` subtree only — it does not clear `content_types/`. If `startTestMigration` can run more than once against the same `current_test_stack_id` (e.g. delta iterations, or re-running a test migration), `schema.json` for that stack would accumulate duplicate content-type entries.

Could you confirm the test stack's `content_types` dir is fresh on every test run (i.e. a new stack id / wiped dir)? If not, always-create needs an upsert/dedupe in `saveContent` rather than an unconditional push.

api/src/utils/field-attacher.utils.ts
```diff
@@ -45,9 +45,13 @@ export const fieldAttacher = async ({ projectId, orgId, destinationStackId, regi
...
+      }
+      else if (iteration === 1) {
```
umesh-more-cstk: **nit:** the `if (isTest)` and `else if (iteration === 1)` branches invoke `contenTypeMaker` with identical arguments. If they're not expected to diverge, this could collapse to `if (isTest || iteration === 1)` to avoid the duplicated call. Non-blocking.

umesh-more-cstk (commented):
## Review: fix test-migration content-type creation + Save-button lock

Non-gating **COMMENT** review. Two independent fixes; both are on the right track and well scoped (touches only `api` + `ui`, matching the Affected Areas — no lockfile/cross-package churn). I traced the field-attacher call sites, the entry-mapper pagination flow, and the content-type write path at the head commit.

### Verified good
- **`field-attacher.utils.ts`** — the `if (isTest) / else if (iteration === 1) / else` restructuring is correct and mutually exclusive; no double-create. The real (final) migration at `migration.service.ts:885` correctly omits `isTest`, so it defaults to `false` and the iteration/skip logic is preserved. The `isTest: true` wiring at `migration.service.ts:453` (test flow) is right.
- `_canSelect = !!entry.contentstackEntryUid` semantics are consistent with the checkbox `rowSelectCheckboxProp` and the Save intent.

### Blocking
- **`entryMapper.tsx:698`** — the new Save-disable predicate inspects only the current page (`tableData`), which for this step is a single server-paginated slice. It can disable Save when the content type *does* have mappable entries (they're on another page) or when the user has unsaved selections made on a previous page. Details + fix inline.

### Non-blocking
- **`field-attacher.utils.ts:50-51`** (question) — always-create on test migrations interacts with the append-based `saveContent` (`schemaData.push`) and `clearStaleEntries` (scoped to `entries/` only, so `content_types/schema.json` is never cleared). Please confirm the test stack / its `content_types` dir is fresh per run so repeated/delta test migrations don't accumulate duplicate content-type schema entries.
- **`field-attacher.utils.ts:53`** (nit) — the `isTest` and `iteration === 1` branches call `contenTypeMaker` with byte-identical args; could collapse to one condition.

### Out-of-diff note (PR description)
The description's rationale — *"Previously the isTest block and the iteration block both ran on the same pass, causing each content type to be created twice"* — doesn't match the diff base: the pre-change code had only `if (iteration === 1)`, no `isTest` block, so there was no double-create on `dev`. The actual prior bug this fixes is that **delta (iteration 2+) test migrations skipped content-type creation** into the fresh test stack. The resulting code is correct; just worth aligning the description so reviewers test the right scenario.

ui/src/components/ContentMapper/entryMapper.tsx
yashin4112: resolved.

api/src/utils/field-attacher.utils.ts
yashin4112: resolved.

api/src/utils/field-attacher.utils.ts
yashin4112: resolved.

vikrantraut-cstk (approved):

shradha-nahar (approved):

PR #1122, opened 2026-07-27
CMG-1093 | Contentful migration got stuck some time

chetan-contentstack:
## 🔗 Jira Ticket

[CMG-1093 — Delta migration | v1.0.0 | Contentful migration got stuck some time](https://contentstack.atlassian.net/browse/CMG-1093)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- `writeOneFile` now uses `fs.promises.writeFile` and is properly awaited. It was previously calling the callback-based `fs.writeFile` in a fire-and-forget pattern, so the outer `await writeOneFile(...)` returned before the file was actually written to disk.
- In `createAssets`, added `fs.promises.mkdir(assetsSave, { recursive: true })` before writing the assets schema/failed files, so the destination directory is guaranteed to exist.

### Why?

During delta Contentful migrations the process would occasionally hang or leave a partial state. Root cause: the assets step kicked off a non-awaited write and moved on, and if the `assetsSave` directory did not yet exist the callback error was swallowed via `console.error`, leaving downstream steps waiting on a file that was never written. Both issues are fixed here.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [ ] `ui` — React frontend
- [ ] Other:

---

## 🧪 How to Test

1. Run a Contentful delta migration on a project that produced the hang in v1.0.0 (any project with assets works).
2. Watch the API logs and the `assets/` output directory as the assets step runs.
3. Confirm the migration proceeds past the assets step without stalling, and that `assets/<project>/assets.json` (and the failed-assets file, if any) are written before the next step begins.

**Expected result:** Migration completes end-to-end; the assets directory is created if missing, `assets.json` is fully written before subsequent steps run, and no `Error writing file: 3` messages appear in logs.

---

## 📸 Screenshots / Recordings

N/A — backend-only fix, no UI changes.

---

- None

---

## ✅ Author Checklist

- [x] Branch follows naming convention: `bugfix/cmg-1093-contentful-migration-issue`
- [ ] `.env` / `example.env` updated if new environment variables were added — N/A
- [ ] New tests written (or not applicable — no unit test exists for this file path and the fix is a race-condition/IO ordering change best verified via the manual repro above)
- [ ] `README.md` / docs updated if behaviour changed — N/A

---

## 👀 Reviewer Notes

Please focus on:
- The `writeOneFile` change — confirm no other caller relies on the previous non-awaiting behaviour (`grep` shows all call sites already `await` it).
- The `mkdir` placement in `createAssets` — it runs after `Promise.all(tasks)` and before the first `writeOneFile` in that block, so both the schema file and the failed-assets file land in an existing directory.

### DIFF
--- a/api/src/services/contentful.service.ts
+++ b/api/src/services/contentful.service.ts
@@ -830,6 +830,7 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
       err
     )
     await customLogger(projectId, destination_stack_id, 'error', message);
+    throw err;
   }
 };
 
--- a/api/src/services/contentful.service.ts
+++ b/api/src/services/contentful.service.ts
@@ -335,11 +335,7 @@ async function readFile(filePath: string, fileName: string) {
  * @throws {Error} - If there is an error writing the file.
  */
 async function writeOneFile(indexPath: string, fileMeta: any) {
-  fs.writeFile(indexPath, JSON.stringify(fileMeta), (err) => {
-    if (err) {
-      console.error("Error writing file: 3", err);
-    }
-  });
+  await fs.promises.writeFile(indexPath, JSON.stringify(fileMeta));
 }
 
 /**
@@ -798,6 +794,7 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
       );
 
       await Promise.all(tasks);
+      await fs.promises.mkdir(assetsSave, { recursive: true });
       const assetMasterFolderPath = path.join(assetsSave, ASSETS_FAILED_FILE);
 
       await writeOneFile(path.join(assetsSave, ASSETS_SCHEMA_FILE), assetData);
--- a/api/src/controllers/migration.controller.ts
+++ b/api/src/controllers/migration.controller.ts
@@ -25,6 +25,9 @@ const getAuditData = async (req: Request, res: Response): Promise<void> => {
  */
 const startTestMigration = async (req: Request, res: Response): Promise<void> => {
   const resp = migrationService.startTestMigration(req);
+  Promise.resolve(resp).catch((err) => {
+    console.error('startTestMigration failed', err);
+  });
   res.status(200).json(resp);
 };
 
@@ -38,6 +41,9 @@ const startTestMigration = async (req: Request, res: Response): Promise<void> =>
  */
 const startMigration = async (req: Request, res: Response): Promise<void> => {
   const resp = migrationService.startMigration(req);
+  Promise.resolve(resp).catch((err) => {
+    console.error('startMigration failed', err);
+  });
   res.status(200).json(resp);
 };
 


api/src/services/contentful.service.ts
```diff
@@ -335,11 +335,7 @@ async function readFile(filePath: string, fileName: string) {
  * @throws {Error} - If there is an error writing the file.
  */
 async function writeOneFile(indexPath: string, fileMeta: any) {
-  fs.writeFile(indexPath, JSON.stringify(fileMeta), (err) => {
-    if (err) {
-      console.error("Error writing file: 3", err);
-    }
-  });
+  await fs.promises.writeFile(indexPath, JSON.stringify(fileMeta));
```
umesh-more-cstk, line 338: **nit:** Correct fix - verified. One intentional behavior change worth flagging for the record: `writeOneFile` now *rejects* on a write error instead of logging `Error writing file: 3` and swallowing it. I checked all three callers (line 361 inside `writeFile`'s try/catch, lines 800 and 814 inside `createAssets`' try/catch) and they all handle the rejection, so no unhandled promise rejection is introduced. No encoding arg is needed - `fs.promises.writeFile` defaults to utf8 for a string payload.

api/src/services/contentful.service.ts
```diff
@@ -798,6 +794,7 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
       );
 
       await Promise.all(tasks);
+      await fs.promises.mkdir(assetsSave, { recursive: true });
       const assetMasterFolderPath = path.join(assetsSave, ASSETS_FAILED_FILE);
 
       await writeOneFile(path.join(assetsSave, ASSETS_SCHEMA_FILE), assetData);
```
umesh-more-cstk, line 800: **question:** With `writeOneFile` now rejecting on failure, a failed schema write here (or the `index.json`/failed-assets writes at 814/816) unwinds to the `createAssets` catch at ~line 825, which only logs via `customLogger` and does NOT re-throw. `migration.service.ts` then proceeds from `await createAssets(...)` (lines 523/948) straight into `createEntry`, which reads `assets.json`. `readFile` (line 322) returns `undefined` on a missing file rather than hanging, so the entries step likely degrades to broken references rather than a hard stall - but since this ticket is specifically about migrations getting *stuck*, can you confirm the entries step tolerates a missing/partial `assets.json` gracefully? If a write genuinely fails, silently continuing may just move the failure downstream; consider surfacing it (or marking the step failed) rather than log-and-continue.

umesh-more-cstk (commented):
**COMMENT review - non-gating.** The human decides the merge gate.

**Verdict: clean, correct bug fix.** I traced the root cause and both changes and they hold up.

**Verified-good**
- `writeOneFile` (line 338): the old callback-based `fs.writeFile` was fire-and-forget, so `await writeOneFile(...)` resolved *before* the bytes hit disk. `migration.service.ts` awaits `createAssets` and then immediately runs `createEntry` (lines 523 -> 535, and 948), and the entries step reads `assets.json` to resolve asset references - a write/read race that matches the reported "stuck / partial state" symptom. Switching to `await fs.promises.writeFile` fixes it. `writeOneFile` is module-private (not exported), and all three call sites - 361 inside `writeFile`'s try/catch, and 800 & 814 inside `createAssets`' try/catch - already handle a rejection, so the now-throwing behavior introduces no unhandled rejection.
- `mkdir(assetsSave, { recursive: true })` (line 797): necessary and correctly placed. Lines 800 & 814 write via `writeOneFile` directly, which does NOT mkdir (unlike the `writeFile` helper at line 355 used on line 816). `assetsSave` is otherwise only created as a side effect of a *successful* per-asset write (`saveAsset`, line 722, on the `files/<id>` subdir). If every asset fails or throws before any success, `assetsSave` would not exist and the two direct `writeOneFile` writes would ENOENT - exactly what this guards.

**Non-blocking**
- _question_ - error swallowing after the fix: see inline on line 800.
- _nit_ - no automated test accompanies an IO-ordering fix (author acknowledged this in the checklist). Given the race nature, a test asserting `assets.json` exists once `createAssets` resolves would lock in the fix and prevent regression.

**Scope:** diff is a single file in `api`, matching the stated Affected Area and the ticket. No unrelated churn.

api/src/services/contentful.service.ts
chetan-contentstack: Thanks — good catch on flagging the behavior change explicitly. Confirmed: all three callers (line 361 inside writeFile's try/catch, and 800/814 inside createAssets' try/catch) already handle the rejection, so no unhandled promise rejection. No change needed.

api/src/services/contentful.service.ts
chetan-contentstack: Addressed in the follow-up commit. createAssets now re-throws after logging, so a failed write (or any asset-step error) propagates to the caller in migration.service.ts instead of silently degrading createEntry into unresolved references. Kept the log line so we still get the structured customLogger entry before

api/src/services/contentful.service.ts
```diff
@@ -833,6 +830,7 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
       err
     )
     await customLogger(projectId, destination_stack_id, 'error', message);
+    throw err;
```
umesh-more-cstk, line 833: **blocker:** Re-throwing here is the right *intent* (abort the sequential step chain so `createEntry` never runs on a half-written `assets.json`), but trace where the rejection ends up:

- `createAssets` is awaited inside the Contentful branch of `startTestMigration` / `startMigration` (`migration.service.ts` L549 / L996), which have **no** outer try/catch.
- The controller calls them **fire-and-forget**: `const resp = migrationService.startTestMigration(req); res.status(200).json(resp);` — no `await`, no `.catch()` (`migration.controller.ts:27`, and `:40` for `startMigration`). `asyncRouter` only catches the *controller's* own rejection, not this detached promise.
- There is no `process.on('unhandledRejection')` anywhere in `api` and no `--unhandled-rejections` flag in the start scripts.

So any FS failure the outer `catch` sees (this `readFile`, `JSON.parse`, `mkdir`, or a `writeOneFile` reject) now becomes an **unhandled promise rejection**. On Node's default mode (`throw`, v15+) that terminates the entire API process — killing all other in-flight migrations — and the user already received `200`, so the UI never sees the failure. Previously the swallow-and-continue kept the process alive.

Suggested fix: don't let it propagate to the detached promise. Either handle it where the migration actually runs, e.g. in `startTestMigration`/`startMigration` wrap the step switch and mark the project failed + log (fail fast without crashing), or attach a `.catch(err => { logger.error(err); /* mark failed */ })` at the fire-and-forget call in the controller, and/or register a global `process.on('unhandledRejection', ...)` as a safety net. `customLogger` already logs here, so the throw's job is just the abort — it needs a catcher upstream.

api/src/services/contentful.service.ts
```diff
@@ -798,6 +794,7 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
       );
 
       await Promise.all(tasks);
+      await fs.promises.mkdir(assetsSave, { recursive: true });
```
umesh-more-cstk, line 797: **nit:** (verified good) Correct and necessary. When every asset fails, `saveAsset` only `mkdir`s `assetPath` on its success path, so `assetsSave` may not exist when `writeOneFile(assets.json)` runs a couple lines below — this `recursive` mkdir prevents that `ENOENT`, and is idempotent for the normal path. No change requested.

umesh-more-cstk (commented):
### Re-review at head `86c00d7d` (COMMENT — non-gating)

Re-reviewed after the new commit `86c00d7d` *"fix(contentful): rethrow from createAssets so write failures surface"* landed on top of the previously-reviewed `0757fd0e`. The only change since the last review is the single added line `throw err;` in the `createAssets` catch block. The diff is still confined to `api/src/services/contentful.service.ts` (3 hunks) — scope is clean and matches the `api` Affected Area.

**Status of the three points from the prior review (`0757fd0e`):**
1. *nit — `writeOneFile` now rejects instead of swallowing.* Resolved by design. All three call sites (`createAssets` L800/L814 inside its `try`, and `writeFile` L361 inside its own `try`) are covered by a `catch`, and the new `throw err` is exactly what makes a `writeOneFile` rejection surface. Correct.
2. *question — does downstream `createEntry` tolerate a missing/partial `assets.json` after a failed write?* Effectively addressed **for the same run**: `startTestMigration`/`startMigration` run the Contentful steps as a sequential `await` chain, so a throw from `createAssets` now aborts before `createTaxonomy`/`createEntry`/`createVersionFile` run — `createEntry` is no longer reached with a half-written `assets.json`. See the blocker below for the caveat on *how* that abort surfaces.
3. *nit — no accompanying test.* Still not addressed; no test file in the diff. Non-blocking (see body note).

---

#### Blocking
- **`contentful.service.ts:833` — the propagated error lands on a fire-and-forget promise with no handler.** The controller invokes `migrationService.startTestMigration(req)` / `startMigration(req)` **without `await` and without `.catch()`** (`api/src/controllers/migration.controller.ts:27` and `:40`), responds `200` immediately, and the migration runs detached. There is **no global `process.on('unhandledRejection')`** anywhere in `api` and no `--unhandled-rejections` flag in the start scripts. So the new `throw err` turns an asset write/`mkdir`/read failure into an **unhandled promise rejection**, which on Node's default (`throw`, v15+) terminates the whole API process — taking down every in-flight migration — while the client already got a `200` and never learns it failed. Before this PR, `createAssets` swallowed the error and continued. Details/fix options inline.

#### Non-blocking
- **No test (prior nit #3, still open).** The author's checklist rationale (race-condition/IO fix) is now weaker: with the `await` + `throw`, the behavior is deterministic and unit-testable (e.g. stub `fs.promises.writeFile` to reject and assert `createAssets` rejects; stub it to resolve and assert `assets.json` content). A small test would lock in the fix.

#### Verified good
- **`writeOneFile` `await fs.promises.writeFile` (L338)** — the previous callback form genuinely resolved the outer `await` before the write completed; this is a real fix.
- **`mkdir(assetsSave, { recursive: true })` (L797)** — necessary and correct: when every asset fails, `saveAsset` never creates a dir under `assetsSave` (it only `mkdir`s `assetPath` on the success path), so the subsequent `writeOneFile(assets.json)` would `ENOENT` without this. `recursive` makes it idempotent for the normal path.
- The failure is still logged via `customLogger` immediately before the throw, so it is not silent in the logs — the gap is purely the missing top-level handler.

#### Out-of-diff notes (can't anchor inline)
- The fire-and-forget controller (`migration.controller.ts:27`/`:40`) and the absence of a global rejection handler are pre-existing and systemic — `aem.service.ts` `createAssets` already propagates (no outer try/catch), so this PR brings Contentful in line with that pattern rather than inventing it. That's why the recommended fix likely belongs in `startTestMigration`/`startMigration` (wrap the switch, mark the project failed, log) or in the controller, not necessarily in this file.

vikrantraut-cstk (approved):

vikrantraut-cstk (approved):

umesh-more-cstk (approved):

PR #1128, opened 2026-07-31
CMG-1101: fix delta reference resolution and locale-migrated tracking

chetan-contentstack:
## 🔗 Jira Tickets
Parent story: [CMG-789 — Delta Migration | v1.0.0 | Update Feature Implementation](https://contentstack.atlassian.net/browse/CMG-789)

- [CMG-1101 — Contentful delta: reference lost & new locale not populated on iteration 2+](https://contentstack.atlassian.net/browse/CMG-1101)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- **CMG-1101 — References resolve on the delta/update path.** The master-locale bulk import resolves references via the CLI's uid-mapper; the update path skipped it and wrote source-cms ids straight to the destination. Threads the flat + per-locale entry uid-mapper into the update config as `__entryMapping__` and remaps Link(Entry) uids in `entry-update-script.cjs`.
- **CMG-1101 — Only actually-processed locales get marked "migrated".** Was recording all configured locales after every run, so a newly-added locale got flagged done before it ran and was skipped on later restarts. Iteration 1 keeps the old behaviour (single bulk); iteration 2+ records master + locales present in `updated-entries.json`.
- **Typo in `contentful.service.ts:439`:** `Array?.isArray(entryId?.id)` → `Array.isArray(entryId?.[id])`.
- Unit coverage added for the new helpers.

### Why?

Reproduced with the bundled `contentful-export.json` / `-v2.json` / `-v3.json`: on iteration 2 the destination reference field went empty and the newly-added locale had zero entries.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [ ] `ui`
- [ ] `upload-api`

---

## 🧪 How to Test

Use the three bundled fixtures.

1. **Iter 1** — migrate `contentful-export.json` (1 locale, 2 refs). Confirm both refs resolve in destination.
2. **Iter 2** — restart with `contentful-export-v2.json` (adds `en-IN`, updates one entry's title). On Map Entry, select the changed entry. After migration, confirm refs still resolve in `en-US` **and** `en-IN` is populated.
3. **Iter 3** — restart with `contentful-export-v3.json` (adds `en-GB`, adds `relatedArticles` ref). Confirm both `author` and `relatedArticles` resolve in all three locales.
4. `cd api && npm test` passes.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
| Iter 2: `author` ref empty in destination after re-mapping | Ref resolves to destination uid |
| Iter 2: `en-IN` locale has 0 entries | `en-IN` populated for every entry |
| Iter 3: `relatedArticles` empty across locales | Both refs resolve in all 3 locales |

---

- Parent story: [CMG-789](https://contentstack.atlassian.net/browse/CMG-789)

---

## ✅ Author Checklist

- [x] Branch follows naming convention: `bugfix/cmg-1101-delta-ref-locale`
- [x] Jira ticket linked
- [x] Self-reviewed the diff
- [x] No secrets committed
- [x] Existing tests pass (`npm test` in `api/`)
- [x] New tests written for the added helpers

---

## 👀 Reviewer Notes

- Reference resolution precedence: per-locale new → per-locale old → flat new → flat old → identity. Identity fallback preserves current behaviour for non-Contentful sources.
- Iteration-1 locale tracking is intentionally unchanged — that path is a single bulk import, so the full configured set is what actually ran.
- No changes outside `api/src/` and `api/tests/`.


api/src/services/runCli.service.ts
```diff
@@ -322,20 +323,54 @@ export const runCli = async (
...
+            }
+          }
+          ranLocales = Array.from(
+            new Set([
+              ...Object.keys(proj?.master_locale ?? {}),
+              ...extractLocalesFromUpdateConfig(updateConfig),
+            ]),
+          );
+        }
+        await recordMigratedLocales(projectId, ranLocales);
```
umesh-more-cstk: **blocker:** This records the locales as migrated *before* the step that actually migrates them.

Order in `startMigration` (migration.service.ts:1243-1282) is: `removeEntriesFromDatabase` writes `updated-entries.json` → `await utilsCli.runCli(...)` (**this line runs here**) → `enrichConfigWith*` → `await utilsUpdateCli.updateEntryCli(...)`. The locales computed above come from `updated-entries.json`, i.e. precisely the entries the *later* `updateEntryCli` step localizes. And `updateEntryCli` swallows all failures (`updateEntryCli.service.ts:240-249` only logs), so the caller cannot tell it failed.

Failure scenario: locale `fr-fr` is added on iteration 2, the CLI import succeeds, `fr-fr` is recorded here, then the update/localize CLI dies (bad auth, rate limit, `entry.update` rejection). `migrated_locales` now contains `fr-fr`, so on the next restart `isFullMigrationForLocale` → false → `isNewLocaleInDelta` → false → only entries the user explicitly marked `isUpdate` get localized. The rest of `fr-fr` is never localized again — the exact permanent-skip bug this PR is fixing.

Suggested fix: move the recording out of `runCli` into `startMigration`, after the update CLI resolves, and make that step surface failure:
```ts
await utilsUpdateCli?.updateEntryCli(/* ... */);   // rethrow on failure
await recordMigratedLocales(safePid, ranLocales);
```
For iteration 1 (no delta/localize step) recording inside `runCli` is fine — it's only the `currentIteration > 1` branch that needs to move.

api/src/services/runCli.service.ts
```diff
@@ -322,20 +323,54 @@ export const runCli = async (
...
+            try {
+              updateConfig = JSON.parse(fs.readFileSync(updatedEntriesPath, 'utf-8'));
+            } catch {
+              updateConfig = null;
+            }
+          }
+          ranLocales = Array.from(
+            new Set([
+              ...Object.keys(proj?.master_locale ?? {}),
+              ...extractLocalesFromUpdateConfig(updateConfig),
```
umesh-more-cstk: **question:** `updated-entries.json` only contains entries that already had a Contentstack uid — `removeEntriesFromDatabase` does `if (!csEntryUid) continue;` (entry-update.utils.ts:188-190) and leaves brand-new entries in the import data to be created, without adding them to `entriesToUpdate`.

So if a locale is added on iteration 2 **and** every source entry for it is new (e.g. a content type first selected in this same iteration), no `…::<locale>` key is ever written, `extractLocalesFromUpdateConfig` returns `[]`, and the locale is fully imported by the CLI but never recorded as migrated. `isFullMigrationForLocale` then stays `true` for it forever, so on every later restart `isNewLocaleInDelta` is `true` and *all* of that locale's entries get force-routed through the localize path, ignoring the user's `isUpdate` selection.

Worth unioning in the locales that this run's import actually touched, not just the ones it queued for update — e.g. the locale directory names under the migration-data entries tree, or `Object.keys(uid-mapper.entryByLocale)` for this iteration (already written by `writePerLocaleEntryUidMapping` a few lines earlier). Is this case in scope for the ticket, or knowingly deferred?

api/src/utils/entry-update.utils.ts
```diff
@@ -297,6 +297,64 @@ export const enrichConfigWithAssetMapping = (
...
+    loggerPath?: string
+): void => {
+    const dbBase = path.join(process.cwd(), DATABASE_FILES.DIRECTORY, projectId);
+
+    const readEntryMapper = (iter: number): { flat: Record<string, string>; byLocale: Record<string, Record<string, string>> } => {
+        const p = path.join(dbBase, iter.toString(), DATABASE_FILES.UID_MAPPER);
+        if (!fs.existsSync(p)) return { flat: {}, byLocale: {} };
+        try {
+            const data = JSON.parse(fs.readFileSync(p, "utf-8"));
+            return { flat: data?.entry || {}, byLocale: data?.entryByLocale || {} };
```
umesh-more-cstk: **question:** `data?.entry` is consumed here as a flat `sourceUid → destUid` map, but the other reader of this exact file/key in the codebase does not assume that shape. `contentMapper.service.ts:2206-2233` (`getEntryUidMap` + `flattenNestedUidMap`) explicitly handles `entry` being **nested per content type** (`{ [ct]: { [src]: dest } }`) and also merges in a `data.entryUid` variant. `models/uidMapper.ts` types it as `Record<string, any>`, and `mergeUidMaps` in `uid-mapper.utils.ts:9-14` documents the same duality ("plain strings … or one-level nested objects (per-content-type entry maps)").

If the nested shape shows up, every `entryMapping.*.flat[sourceUid]` lookup in `resolveReferenceUid` misses and silently falls through to the identity fallback — i.e. the bug this PR fixes reappears whenever the per-locale map has no hit. Worse, if a source uid ever collides with a content-type key, `flat[sourceUid]` is an object and gets written straight into `uid`.

Suggest reusing the existing normalisation rather than duplicating the assumption — extract `flattenNestedUidMap` into a shared util and do:
```ts
return {
  flat: flattenNestedUidMap({ ...(data?.entry ?? {}), ...(data?.entryUid ?? {}) }),
  byLocale: data?.entryByLocale || {},
};
```

api/src/utils/entry-update-script.cjs
```diff
@@ -4,6 +4,63 @@ const isAssetField = (value) =>
     value && typeof value === 'object' && !Array.isArray(value) &&
     'urlPath' in value && 'filename' in value;
 
+/** Shape produced by processField's 'reference' case: { uid, _content_type_uid }. */
+const isReferenceValue = (value) =>
+    value && typeof value === 'object' && !Array.isArray(value) &&
+    'uid' in value && '_content_type_uid' in value;
+
+const isReferenceArray = (value) =>
+    Array.isArray(value) && value.length > 0 && value.every(isReferenceValue);
```
umesh-more-cstk: **nit:** `every(isReferenceValue)` makes this all-or-nothing, and the producers can legitimately emit mixed arrays:

- `processArrayFields` (contentful.service.ts:543-555) pushes the **original raw Contentful link object** when the target id isn't in the references map ("Keep unresolved references intact").
- the single-reference path (contentful.service.ts:443) returns `[entryId?.[id]]`, i.e. `[undefined]`, when the target wasn't exported.

One such element makes `isReferenceArray` false, so `resolveReferenceField` is never called and *every* resolvable uid in that multi-reference field keeps its source id. Since `resolveReferenceField` already maps per item, loosening the gate to `.some()` and passing non-reference items straight through is strictly better:
```js
const isReferenceArray = (value) =>
    Array.isArray(value) && value.length > 0 && value.some(isReferenceValue);
// in resolveReferenceField:
return value.map((item) => (isReferenceValue(item)
    ? { ...item, uid: resolveReferenceUid(item.uid, locale, entryMapping) }
    : item));
```

api/src/utils/entry-update-script.cjs
```diff
@@ -89,6 +146,8 @@ const mergeFlatPayloadIntoEntry = async (entry, entryUid, updateData, oldMapping
                 oldMapping,
                 newMapping
             );
+        } else if (isReferenceValue(nextVal) || isReferenceArray(nextVal)) {
```
umesh-more-cstk, line 158: **question:** Resolution is top-level-field only. References that live *inside* a value are missed:

- group / modular-block values whose nested arrays were built by `processArrayFields` (contentful.service.ts:537-560) — the outer value is a plain object, so `isReferenceValue` is false and the whole subtree is skipped;
- entries embedded in a JSON RTE (`processRTEOrNestedObject` → `jsonRTE`, primed from `rteRefrences`, which `createRefrence` also fills with **source** ids at contentful.service.ts:1591-1595).

Those keep the source uid on localized entries, i.e. the reported symptom persists for any content model that nests its references. Was that scoped out deliberately for this ticket? A small recursive walk (descend into plain objects/arrays, apply the same two predicates at each level) would cover both without changing the top-level behaviour or the tests.

api/src/services/contentful.service.ts
```diff
@@ -436,7 +436,7 @@ const processField = (
         return refs;
       }
       const id = lang_value?.sys?.id;
-      if(Array?.isArray(entryId?.id)){
+      if(Array.isArray(entryId?.[id])){
```
umesh-more-cstk, line 439: **nit:** The typo fix is right (`entryId?.id` was reading a property literally named `id`), but the branch it repairs is unreachable, so net behaviour is unchanged: `createRefrence` builds `refrences[id] = { uid, _content_type_uid }` (line 1584) — values are always objects, never arrays — so `Array.isArray(entryId?.[id])` is always false and the `else` (`[entryId?.[id]]`) always runs, exactly as before. Fine as a cleanup; just worth not expecting a behaviour change from it.

The real gap on this path is left untouched: when the target id isn't in the references map (reference to an entry that wasn't exported), or when `lang_value` is an array while `fieldData.advanced.multiple` is falsy (so `lang_value?.sys?.id` is `undefined`), this returns `[undefined]`. That both ships a broken value into the import data and — after this PR — defeats the new `isReferenceArray` resolution in `entry-update-script.cjs`. Suggest:
```ts
const id = lang_value?.sys?.id;
const mapped = entryId?.[id];
if (!mapped) return [];
return Array.isArray(mapped) ? mapped : [mapped];
```

package.json
```diff
@@ -46,8 +46,8 @@
     "@tootallnate/once": ">=3.0.1",
     "fast-xml-parser": ">=5.3.8",
     "diff": ">=5.2.2",
-    "fast-uri": ">=3.1.2",
-    "brace-expansion": ">=5.0.6"
+    "fast-uri": ">=4.1.1",
+    "brace-expansion": ">=5.0.9"
```
umesh-more-cstk, line 50: **blocker:** This bump can't clear the Snyk SLA breach for the deployables. Root `package.json` declares no `workspaces` and its only dependencies are `husky` + `validate-branch-name` (the run scripts `cd ./api`, `cd ./ui`, `cd ./upload-api` — those are separate installs with their own lockfiles), so these `overrides` govern almost nothing at runtime.

The actual services still resolve the vulnerable versions at this commit:

| project | fast-uri | brace-expansion |
|---|---|---|
| `api/package-lock.json` | **3.1.2** | **5.0.6** |
| `ui/package-lock.json` | — | **5.0.6** |
| `upload-api/package-lock.json` | **3.1.2** | **5.0.7** |

None of `api|ui|upload-api/package.json` override `fast-uri` or `brace-expansion` (they pin `ajv`/`minimatch`, which is what pulls these in transitively). Fix: add `"fast-uri": ">=4.1.1"` / `"brace-expansion": ">=5.0.9"` to each affected sub-project's `overrides` and regenerate those lockfiles — otherwise Snyk will keep reporting them.

Separately, on scope: this bump (commit `cb7c1ef1`) is unrelated to the delta and Contentful fixes in the same branch. Splitting it out would make both easier to review and to revert independently.

umesh-more-cstk (commented):
## Review summary

Three independent changes here: (a) a delta-migration fix that threads entry uid-mapper data into the entry-update script so reference fields resolve to real Contentstack uids, (b) a change to how `migrated_locales` is recorded, and (c) a root-level dependency override bump. The delta reference-resolution work is sound and well tested — I traced the premise end-to-end and it holds: `removeEntriesFromDatabase` snapshots the on-disk transformed entries *before* the CLI import, and `createRefrence` (contentful.service.ts:1584) writes the **source** cms id into `{ uid, _content_type_uid }`, so this update path really was writing source-side uids onto localized entries. Wiring order is also correct — `writeUidMapping`/`writePerLocaleEntryUidMapping` run inside `runCli` before `enrichConfigWithEntryMapping` is called at migration.service.ts:1265.

### Blocking
1. **runCli.service.ts:373** — `recordMigratedLocales` still runs *before* the localize/update CLI that actually writes those locales, and that CLI swallows every failure. Same premature-marking class of bug this PR is fixing, just via a different trigger.
2. **package.json:49-50** — the Snyk bump only touches the root project; the three deployables still pin the vulnerable versions.

### Non-blocking
3. **entry-update.utils.ts:328** — `data?.entry` is read as flat, but the sibling reader in `contentMapper.service.ts` explicitly handles a nested-per-content-type shape.
4. **runCli.service.ts:369** — a new locale whose entries are *all* newly created never lands in `updated-entries.json`, so it is never recorded as migrated.
5. **entry-update-script.cjs:13** — `isReferenceArray` is all-or-nothing; one unresolved element disables resolution for the whole field.
6. **entry-update-script.cjs:149** — resolution is top-level only (group/modular-block/JSON-RTE-embedded references still keep source uids).
7. **contentful.service.ts:439** — the typo fix is correct but the branch it fixes is unreachable, so net behaviour is unchanged.

### Verified good
- `enrichConfigWithEntryMapping` mirrors `enrichConfigWithAssetMapping` exactly (try/catch, empty-mapping default, non-throwing on a malformed config) and never blocks the main flow.
- `extractLocalesFromUpdateConfig`'s `__`-prefix skip correctly excludes `__assetMapping__` / `__entryMapping__` / `__assetUpdates__`, and it was verified that those keys are written *after* `runCli` reads the file — so the skip is belt-and-braces, not load-bearing.
- Locale-code direction is consistent on both sides: `updated-entries.json` keys carry the **destination** code (`removeEntriesFromDatabase` uses the on-disk locale dir name), which is what `migrated_locales` / `isFullMigrationForLocale` compare against.
- New unit tests are genuinely good — they cover the full fallback precedence chain, the back-compat identity path with no mapping supplied, and malformed/absent config files.

### Out-of-diff notes (can't be anchored inline)
- `api/src/services/updateEntryCli.service.ts:240-249` — the `catch` only logs; a failed entry-update CLI run is indistinguishable from success to the caller. This is what makes finding #1 a silent data problem rather than a retryable one.
- `api/src/services/contentMapper.service.ts:2206-2233` — `getEntryUidMap` / `flattenNestedUidMap`, the existing precedent for reading `uid-mapper.json`'s `entry` key (handles nested shape *and* an `entryUid` variant). Relevant to finding #3.
- `api/src/utils/entry-update.utils.ts:188-190` — `if (!csEntryUid) continue;` is why brand-new entries never reach `entriesToUpdate`. Relevant to finding #4.
- Lockfile evidence for finding #2: `api/package-lock.json` fast-uri **3.1.2** / brace-expansion **5.0.6**; `ui/package-lock.json` brace-expansion **5.0.6**; `upload-api/package-lock.json` fast-uri **3.1.2** / brace-expansion **5.0.7**. Root `package.json` declares no `workspaces`, and its only deps are husky + validate-branch-name.
- Process note: I could not read the PR description / PR-type / Affected-Areas checkboxes (the GitHub API is SAML-SSO blocked for my token), so I could not check the diff against the stated scope. Flagging on content alone: the dependency bump commit (`cb7c1ef1`) is unrelated to the delta and Contentful fixes in the same branch and would be easier to review — and to revert — on its own.

umesh-more-cstk:
Follow-up to my review above — the SAML-SSO block on my token is resolved, so I could finally read the PR description and check the diff against the stated scope. Retracting the "could not check" caveat at the end of that review; here is the actual result.

The three functional bullets under **What changed?** match the diff accurately, and **Affected Areas → `api`** is right for them. The dependency bump is the part that isn't covered:

- **PR Type** has only `🐛 Bug Fix` ticked. Commit `cb7c1ef1` bumps the `fast-uri` and `brace-expansion` overrides (`package.json` +2/-2, `package-lock.json` +7/-7), so `🧹 Chore / Dependency Update` applies too.
- **What changed?** does not mention the bump at all — a reader reviewing from the description alone would not know it is in here.
- **Affected Areas** — the bump lives in the **root** `package.json`, which none of the listed areas covers. Worth noting that `ui` and `upload-api` being unticked is consistent with blocker #2: the override genuinely does not reach them, which is exactly the problem.
- **How to Test** covers the three migration iterations and `cd api && npm test`, but has no step that verifies the vulnerable versions actually resolved away. Given blocker #2, a `npm ls fast-uri brace-expansion` in each of `api`, `ui`, `upload-api` would have caught it before review.

None of this changes the technical findings above — the two blockers and five non-blocking items stand as written.

vikrantraut-cstk (dismissed):

PR #1136, opened 2026-08-05
Bugfix(delta) | References, Entry Mapping, RTE, empty-state, CT count

chetan-contentstack:
## 🔗 Jira Tickets

Parent story: [CMG-789 — Delta Migration | v1.0.0](https://contentstack.atlassian.net/browse/CMG-789)

- [CMG-1101 — Iter 2 reference removed on entries](https://contentstack.atlassian.net/browse/CMG-1101)
- [CMG-1102 — asic_gender updated entries not shown on Map Entry](https://contentstack.atlassian.net/browse/CMG-1102)
- [CMG-1097 — Assets screen empty-state layout broken](https://contentstack.atlassian.net/browse/CMG-1097)
- [CMG-1103 — JSON RTE hyperlink URL lost in destination](https://contentstack.atlassian.net/browse/CMG-1103)
- [CMG-1104 — Entry Mapper selection lost across locales](https://contentstack.atlassian.net/browse/CMG-1104)
- [CMG-1095 — Delta iter 2+: not all content types displayed on Map Entry](https://contentstack.atlassian.net/browse/CMG-1095)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [x] 🧹 Chore / Dependency Update
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- **CMG-1101** — Entry uid-mapper (flat + per-locale) threaded into the update config so `entry-update-script.cjs` resolves Link(Entry) fields on the delta/localize path; locale-migrated tracking moved after `updateEntryCli` resolves so silent failures don't permanently flag locales as done.
- **CMG-1102** — `extractEntries.js` now uses each content type's `displayField` with fallbacks (title → name → first localized string → sys.id) instead of hard-coded `title`/`name`, so entries with other display fields no longer disappear from Map Entry.
- **CMG-1097** — Asset mapper search-empty state: force `.Table` height when it contains an `.EmptyState` and center `.Table__centerWrapper` so the illustration + heading + description render properly.
- **CMG-1103** — JSON RTE hyperlinks rewritten to Contentstack's shape: plain `→ type:'a' attrs.url`, entry/asset `→ type:'reference'` with `display-type:'link'`. URLs now survive to destination.
- **CMG-1104** — Entry-mapper selection persists across locales: locale-scoped server toggle, stale-fetch generation guard, unified data-load effect, `tableRevision` remount so Venus's Table re-reads `initialSelectedRowIds`.
- **CMG-1095** — `getContentTypes` now unions content types from every prior iteration (1..N-1) instead of only N-1, so types migrated in earlier iterations but absent from N-1 are still classified as "old" on iteration N.
- Dep bumps to clear Snyk SLA (`fast-uri`, `brace-expansion`) in root and each sub-project package.
- Path-traversal SAST fixes on new file reads.

### Why?

Six independent delta-migration defects reported against v1.0.0. All share the same delta code paths so grouping keeps the review context tight.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [x] `ui` — React frontend
- [x] `upload-api` — Upload API server
- [x] Other: root `package.json` (dep override bumps)

---

## 🧪 How to Test

Fresh project on `bugfix/cmg-1101-ref-locale`. Migrate 3 iterations using the bundled fixtures.

1. **Iter 1** — `contentful-export.json` (3 CTs, 1 locale). Complete migration; references resolve in destination.
2. **Iter 2** — `asic.json` (24 CTs incl. `gender`/`moreInformation`/`tip`/`width` which use `internalTitle`). On Map Entry, all 24 should appear with entries. Save selections on multiple locales and verify each persists across locale switches.
3. **Iter 3** — `asic-merged.json` (27 CTs = iter 1's 3 + iter 2's 24). Map Entry should show **26** content types (one CT genuinely has 0 entries).
4. Assets tab → search for something not present → verify empty-state renders full illustration + heading + description below the header.
5. Any entry with a JSON RTE containing a plain hyperlink (e.g. ASIC `disclaimer`) → verify the link is preserved in the destination JSON RTE.
6. `cd api && npm test` passes.

**Expected result:** All 26 content types visible on iter 3 Map Entry; references and JSON RTE hyperlinks intact in destination; per-locale entry selections persist; assets search empty-state layout correct.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
|        |       |

---

- Parent story: [CMG-789](https://contentstack.atlassian.net/browse/CMG-789)

---

## ✅ Author Checklist

- [x] Branch follows naming convention: `bugfix/cmg-1101-ref-locale`
- [x] Jira tickets linked above
- [x] Self-reviewed the diff
- [x] No secrets committed
- [x] Existing tests pass (`npm test` in `api/`)
- [x] New tests written for reference resolution helpers

---

## 👀 Reviewer Notes

- Reference resolution precedence: per-locale new → per-locale old → flat new → flat old → identity fallback.
- CMG-1095 union-of-prior-iterations approach is O(iteration count) lowdb reads — fine for single-digit iterations.
- Dep bumps are in a separate concern from the delta fixes; splitting into its own PR would be cleaner but was kept here to unblock the Snyk SLA check.

---

### DIFF
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -15,6 +15,7 @@ import {
   CONTENT_TYPE_STATUS,
   VALIDATION_ERRORS,
   MIGRATION_DATA_CONFIG,
+  CMS,
 } from '../constants/index.js';
 import logger from '../utils/logger.js';
 import { config } from '../config/index.js';
@@ -35,6 +36,7 @@ import { isDuplicateEntry } from '../utils/entry-duplicate.utils.js';
 import { getSourceLocaleForDestination } from '../utils/locale-migration.utils.js';
 import { loadPreviousAssetMetadata } from '../utils/asset-update.utils.js';
 import { flattenNestedUidMap } from '../utils/uid-mapper.utils.js';
+import { contentfulService } from './contentful.service.js';
 
 
 const idCorrector = ({ id }: { id: string }) => {
@@ -176,11 +178,7 @@ const putTestData = async (req: Request) => {
 
     const uidMapperCurrent = getUidMapperDb(projectId, iteration);
     await uidMapperCurrent.read();
-    let uidMapperPrev: any = null;
-    if (iteration > 1) {
-      uidMapperPrev = getUidMapperDb(projectId, iteration - 1);
-      await uidMapperPrev.read();
-    }
+    const uidMapperPrev: any = iteration > 1 ? await getNearestPriorUidMapper(projectId, iteration) : null;
 
     const mergeEntry = (base: any, incoming: any) => {
       const keep = { ...(base ?? {}) };
@@ -2009,24 +2007,40 @@ const updateEntryStatus = async (req: Request) => {
       .find({ id: projectId })
       .value();
     const iteration = projectData?.iteration || 1;
-    const EntryMapperModel = getEntryMapperDb(projectId, iteration);
-    await EntryMapperModel.read();
-    const foundEntry: EntryMapper[] = [];
     // Rows in entry_mapper are per-(entry × source-locale); each id is unique per row.
     // Also scope the toggle by source-locale as a safety net so a same-id collision
     // (if it ever happens) can't flip a sibling locale's row and clobber the user's
-    // selection state on the other locale.
+    // selection state on the other locale. Only enforced when the row actually carries
+    // a language — legacy rows created before language-tagging existed have none, and
+    // requiring a match against them would make them permanently untoggleable.
     const sourceLocale = locale
       ? getSourceLocaleForDestination(projectData ?? {}, locale)
       : null;
-    await EntryMapperModel.update((data: any) => {
-      data?.entry_mapper?.forEach((entry: any) => {
-        if (!validatedUids.includes(entry?.id)) return;
-        if (sourceLocale && (entry?.language ?? '') !== sourceLocale) return;
-        entry.isUpdate = !entry.isUpdate;
-        foundEntry.push(entry);
+
+    const toggleInModel = async (iter: number): Promise<EntryMapper[]> => {
+      const model = getEntryMapperDb(projectId, iter);
+      await model.read();
+      const matched: EntryMapper[] = [];
+      await model.update((data: any) => {
+        data?.entry_mapper?.forEach((entry: any) => {
+          if (!validatedUids.includes(entry?.id)) return;
+          if (sourceLocale && entry?.language && entry.language !== sourceLocale) return;
+          entry.isUpdate = !entry.isUpdate;
+          matched.push(entry);
+        });
       });
-    });
+      return matched;
+    };
+
+    let foundEntry = await toggleInModel(iteration);
+
+    // Fallback: mirrors getEntryMapping's read-side fallback (contentMapper.service.ts
+    // ~2119-2131) — right after a restart, before iteration N's entry-mapper rows exist,
+    // Map Entry renders rows sourced from iteration N-1. Without this, saving those rows
+    // 404s here even though the user is looking at exactly what the read path showed them.
+    if (!foundEntry.length && iteration > 1) {
+      foundEntry = await toggleInModel(iteration - 1);
+    }
 
     if (foundEntry.length) {
       return {
@@ -2223,6 +2237,29 @@ const lookupContentstackEntryUidFromUidMap = (
   return String(resolved).trim() || undefined;
 };
 
+/**
+ * Loads the nearest prior iteration's uid-mapper model, walking backward from
+ * `iteration - 1` down to 1 — not just `iteration - 1` alone. `writeUidMapping` already
+ * merges each successful run's uid-mapper.json forward from the one before it, so the
+ * nearest prior iteration that actually has a file already carries everything from every
+ * iteration before it; we only need to skip iterations where the file is simply absent
+ * (e.g. a restart that skipped an actual "Start Migration" run for that iteration — see
+ * CMG-1095, the same gap for content types). Without this, a single skipped iteration
+ * permanently breaks uid resolution for every iteration after it.
+ */
+const getNearestPriorUidMapper = async (projectId: string, iteration: number): Promise<any | null> => {
+  for (let i = iteration - 1; i >= 1; i--) {
+    const model = getUidMapperDb(projectId, i);
+    await model.read();
+    const data = model?.data as any;
+    const hasData =
+      Object.keys(data?.entry ?? {}).length > 0 ||
+      Object.keys(data?.assets ?? {}).length > 0;
+    if (hasData) return model;
+  }
+  return null;
+};
+
 const getEntryUidMap = (uidMapperModel: any): Record<string, any> => {
   const d = uidMapperModel?.data ?? {};
   const pick = (x: unknown): Record<string, any> => {
@@ -2256,11 +2293,7 @@ const enrichEntriesWithUidMapper = async (
   const currentModel = getUidMapperDb(projectId, iteration);
   await currentModel.read();
 
-  let prevModel: any = null;
-  if (iteration > 1) {
-    prevModel = getUidMapperDb(projectId, iteration - 1);
-    await prevModel.read();
-  }
+  const prevModel: any = iteration > 1 ? await getNearestPriorUidMapper(projectId, iteration) : null;
 
   return entries?.map((item: any) => {
     if (!item) return item;
@@ -2347,12 +2380,83 @@ const updateAssetStatus = async (req: Request) => {
   }
 };
 
+/**
+ * Re-attempts the download for one asset that failed during the last migration run
+ * (currently CMS Contentful only — the `cs_failed.json` file this reads is written by
+ * contentfulService.createAssets). Only re-stages the asset locally; it lands in the
+ * destination stack on the next migration run (Start Migration), same as any other asset.
+ */
+const retryAssetDownload = async (req: Request) => {
+  const srcFunc = "retryAssetDownload";
+  const projectId = req?.params?.projectId;
+  const assetUid = req?.params?.assetUid;
+
+  if (!assetUid) {
+    return {
+      status: HTTP_CODES?.BAD_REQUEST,
+      data: { message: "Missing assetUid" },
+    };
+  }
+
+  try {
+    await ProjectModelLowdb.read();
+    const projectData: any = ProjectModelLowdb.chain
+      .get("projects")
+      .find({ id: projectId })
+      .value();
+
+    const destinationStackId = projectData?.destination_stack_id;
+    const filePath = projectData?.legacy_cms?.file_path;
+    const cms = projectData?.legacy_cms?.cms;
+
+    if (!destinationStackId || !filePath) {
+      return {
+        status: HTTP_CODES?.BAD_REQUEST,
+        data: { message: "Project is missing a destination stack or source file path." },
+      };
+    }
+    if (cms !== CMS.CONTENTFUL) {
+      return {
+        status: HTTP_CODES?.BAD_REQUEST,
+        data: { message: "Asset retry is only supported for Contentful projects." },
+      };
+    }
+
+    const cleanLocalPath = filePath.replace(/\/$/, '');
+    const result = await contentfulService.retryFailedAsset(
+      cleanLocalPath,
+      destinationStackId,
+      projectId,
+      assetUid,
+    );
+
+    return {
+      status: result.success ? HTTP_CODES?.OK : HTTP_CODES?.BAD_REQUEST,
+      data: result,
+    };
+  } catch (error: any) {
+    logger.error(
+      getLogMessage(
+        srcFunc,
+        "Error occurred while retrying asset download",
+        error
+      )
+    );
+    throw new ExceptionFunction(
+      error?.message || HTTP_TEXTS.INTERNAL_ERROR,
+      error?.statusCode || error?.status || HTTP_CODES.SERVER_ERROR,
+    );
+  }
+};
+
 const getAssetMapping = async (req: Request) => {
   const srcFunc = "getAssetMapping";
   const projectId = req?.params?.projectId;
   const skip: any = req?.params?.skip;
   const limit: any = req?.params?.limit;
   const search: string = req?.params?.searchText?.toLowerCase();
+  // Optional status filter: ?status=failed | missing | ok. Absent/anything else = no filter.
+  const statusFilter = req?.query?.status as string | undefined;
 
   let result: any[] = [];
   let filteredResult = [];
@@ -2385,15 +2489,27 @@ const getAssetMapping = async (req: Request) => {
     }
 
     // Fill missing contentstackAssetUid from uid-mapper (current first, then
-    // the previous iteration) so rows saved before the import resolve later.
+    // the nearest prior iteration that actually has one) so rows saved before
+    // the import resolve later.
     const uidMapperCurrent = getUidMapperDb(projectId, iteration);
     await uidMapperCurrent.read();
-    let uidMapperPrev: any = null;
-    if (iteration > 1) {
-      uidMapperPrev = getUidMapperDb(projectId, iteration - 1);
-      await uidMapperPrev.read();
-    }
-    const enrichedMapping = (assetMapping ?? []).map((item: any) => {
+    const uidMapperPrev: any = iteration > 1 ? await getNearestPriorUidMapper(projectId, iteration) : null;
+
+    // Whether we actually have any uid data to resolve against yet. getUidMapperDb creates
+    // the file with an empty `assets: {}` default, so a fresh iteration directory (visited
+    // right after a restart, before this iteration's CLI import has run and written
+    // writeUidMapping's output) legitimately has none — distinct from "this project simply
+    // has no previously-migrated assets". Also true if any row already carries a
+    // pre-resolved uid from creation time (putTestData resolves it then, see ~line 280).
+    const hasAnyUidData =
+      Object.keys((uidMapperCurrent?.data as any)?.assets ?? {}).length > 0 ||
+      Object.keys((uidMapperPrev?.data as any)?.assets ?? {}).length > 0 ||
+      (assetMapping ?? []).some((item: any) => {
+        const uid = item?.contentstackAssetUid;
+        return uid != null && String(uid).trim() !== '';
+      });
+
+    let uidEnriched = (assetMapping ?? []).map((item: any) => {
       if (!item) return item;
       const existing = item?.contentstackAssetUid;
       if (existing != null && String(existing).trim() !== '') {
@@ -2405,36 +2521,86 @@ const getAssetMapping = async (req: Request) => {
       return resolved ? { ...item, contentstackAssetUid: resolved } : item;
     });
 
+    // Status per row for the UI: 'missing' (no url/upload in the source at all — nothing
+    // to retry), 'failed' (had a source but the last migration run's download attempt
+    // threw — retriable), or 'ok'. Read once per request; cs_failed.json is only written
+    // after an actual migration run, so it's absent (empty status) before that.
+    let failedAssets: Record<string, any> = {};
+    const destinationStackId = projectData?.destination_stack_id;
+    if (destinationStackId) {
+      const failedPath = path.join(MIGRATION_DATA_CONFIG.DATA, destinationStackId, MIGRATION_DATA_CONFIG.ASSETS_DIR_NAME, MIGRATION_DATA_CONFIG.ASSETS_FAILED_FILE);
+      if (fs.existsSync(failedPath)) {
+        try {
+          failedAssets = JSON.parse(fs.readFileSync(failedPath, 'utf-8')) || {};
+        } catch {
+          failedAssets = {};
+        }
+      }
+    }
+    const enrichedMapping = uidEnriched.map((item: any) => {
+      if (!item) return item;
+      if (item?.hasSource === false) {
+        return { ...item, status: 'missing', errorMessage: 'No source file found for this asset — nothing to migrate.' };
+      }
+      const failure = failedAssets?.[item?.otherCmsAssetUid];
+      if (failure) {
+        return { ...item, status: 'failed', errorMessage: failure?.reason_for_error || 'Failed to download this asset during the last migration run.' };
+      }
+      return { ...item, status: 'ok' };
+    });
+
     // Delta migration intent: on iteration 2+ the Assets tab lists ONLY assets that
     // were already migrated in a prior iteration — i.e. those with a Contentstack
     // uid. The user selects which of those to update with the current file's newer
     // version. Brand-new assets in this iteration have no prior uid; they upload
     // automatically during the run and don't need a Map Entry row (nothing to
     // select or update yet). Iteration 1 is untouched — everything is new then.
-    const displayMapping = iteration > 1
+    //
+    // Exception: always surface 'failed'/'missing' rows even without a uid. A brand-new
+    // asset that fails to download NEVER gets a Contentstack uid (it never successfully
+    // migrates), so the has-uid check alone would hide it from view forever — the user
+    // would have no way to discover or retry it.
+    // Only apply the delta filter once we actually have uid data to filter with —
+    // otherwise a race right after restart (this iteration's uid-mapper.json not written
+    // yet) would filter out EVERY row and render an empty tab indistinguishable from "no
+    // previously-migrated assets", which could be mistaken for correct behavior since
+    // CMG-1097 already gives that empty state a legitimate-looking layout.
+    const displayMapping = iteration > 1 && hasAnyUidData
       ? enrichedMapping.filter((item: any) => {
           const uid = item?.contentstackAssetUid;
-          return uid != null && String(uid).trim() !== '';
+          const hasUid = uid != null && String(uid).trim() !== '';
+          return hasUid || item?.status === 'failed' || item?.status === 'missing';
         })
       : enrichedMapping;
 
-    if (!isEmpty(displayMapping)) {
+    // Aggregate counts across the FULL (unpaginated, unsearched) visible set — the banner
+    // needs "3 assets won't migrate" regardless of which page or search term is active.
+    const missingCount = displayMapping.filter((item: any) => item?.status === 'missing').length;
+    const failedCount = displayMapping.filter((item: any) => item?.status === 'failed').length;
+
+    const statusFiltered = statusFilter && ['ok', 'missing', 'failed'].includes(statusFilter)
+      ? displayMapping.filter((item: any) => item?.status === statusFilter)
+      : displayMapping;
+
+    if (!isEmpty(statusFiltered)) {
       if (search) {
-        filteredResult = displayMapping?.filter?.((item: any) =>
+        filteredResult = statusFiltered?.filter?.((item: any) =>
           item?.filename?.toLowerCase().includes(search) ||
           item?.title?.toLowerCase().includes(search)
         );
         totalCount = filteredResult?.length;
         result = filteredResult?.slice(skip, Number(skip) + Number(limit));
       } else {
-        totalCount = displayMapping?.length;
-        result = displayMapping?.slice(skip, Number(skip) + Number(limit));
+        totalCount = statusFiltered?.length;
+        result = statusFiltered?.slice(skip, Number(skip) + Number(limit));
       }
     }
     return {
       status: HTTP_CODES?.OK,
       count: totalCount,
-      assetMapping: result
+      assetMapping: result,
+      missingCount,
+      failedCount,
     };
 
   } catch (error: any) {
@@ -2473,4 +2639,5 @@ export const contentMapperService = {
   updateEntryStatus,
   getAssetMapping,
   updateAssetStatus,
+  retryAssetDownload,
 };
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -2028,7 +2028,7 @@ const updateEntryStatus = async (req: Request) => {
       });
     });
 
-    if (foundEntry) {
+    if (foundEntry.length) {
       return {
         status: HTTP_CODES?.OK,
         data: foundEntry
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -2405,17 +2405,30 @@ const getAssetMapping = async (req: Request) => {
       return resolved ? { ...item, contentstackAssetUid: resolved } : item;
     });
 
-    if (!isEmpty(enrichedMapping)) {
+    // Delta migration intent: on iteration 2+ the Assets tab lists ONLY assets that
+    // were already migrated in a prior iteration — i.e. those with a Contentstack
+    // uid. The user selects which of those to update with the current file's newer
+    // version. Brand-new assets in this iteration have no prior uid; they upload
+    // automatically during the run and don't need a Map Entry row (nothing to
+    // select or update yet). Iteration 1 is untouched — everything is new then.
+    const displayMapping = iteration > 1
+      ? enrichedMapping.filter((item: any) => {
+          const uid = item?.contentstackAssetUid;
+          return uid != null && String(uid).trim() !== '';
+        })
+      : enrichedMapping;
+
+    if (!isEmpty(displayMapping)) {
       if (search) {
-        filteredResult = enrichedMapping?.filter?.((item: any) =>
+        filteredResult = displayMapping?.filter?.((item: any) =>
           item?.filename?.toLowerCase().includes(search) ||
           item?.title?.toLowerCase().includes(search)
         );
         totalCount = filteredResult?.length;
         result = filteredResult?.slice(skip, Number(skip) + Number(limit));
       } else {
-        totalCount = enrichedMapping?.length;
-        result = enrichedMapping?.slice(skip, Number(skip) + Number(limit));
+        totalCount = displayMapping?.length;
+        result = displayMapping?.slice(skip, Number(skip) + Number(limit));
       }
     }
     return {
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -34,6 +34,7 @@ import getUidMapperDb from "../models/uidMapper.js";
 import { isDuplicateEntry } from '../utils/entry-duplicate.utils.js';
 import { getSourceLocaleForDestination } from '../utils/locale-migration.utils.js';
 import { loadPreviousAssetMetadata } from '../utils/asset-update.utils.js';
+import { flattenNestedUidMap } from '../utils/uid-mapper.utils.js';
 
 
 const idCorrector = ({ id }: { id: string }) => {
@@ -2219,17 +2220,6 @@ const getEntryUidMap = (uidMapperModel: any): Record<string, any> => {
   if (nUid > 0) return fromEntryUid;
   if (nEnt > 0) return fromEntry;
   return {};
-};
-
-const flattenNestedUidMap = (raw: Record<string, any>): Record<string, any> => {
-  const keys = Object?.keys(raw ?? {});
-  if (keys?.length === 0) return {};
-  const nested = keys?.every((k) => {
-    const v = raw[k];
-    return v != null && typeof v === 'object' && !Array.isArray(v);
-  });
-  if (!nested) return { ...raw };
-  return keys.reduce<Record<string, any>>((acc, k) => ({ ...acc, ...raw[k] }), {});
 };
 
 /**
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -469,15 +469,29 @@ const getContentTypes = async (req: Request) => {
     );
 
     // Delta migration: from iteration 2 onwards, split content types into new vs old
-    // relative to the previous iteration so Step 3 (field mapping) shows only new types
-    // and Step 4 (entry mapping) shows only already-migrated types. Iteration 1 is untouched.
+    // relative to EVERY prior iteration (1..N-1) so Step 3 (field mapping) shows only
+    // genuinely-never-seen types and Step 4 (entry mapping) shows every already-migrated
+    // type. Comparing only against iteration N-1 misclassified any content type that
+    // was migrated in iteration 1 but absent from the iteration-2 source as "new" on
+    // iteration 3 — sending already-migrated types back to Map Content Fields and
+    // hiding them from Map Entry. Iteration 1 is untouched.
     if (iteration > 1) {
-      const PrevContentTypesMapperModelLowdb = getContentTypesMapperDb(projectId, iteration - 1);
-      await PrevContentTypesMapperModelLowdb.read();
-      const prevContentMapper =
-        PrevContentTypesMapperModelLowdb.chain.get('ContentTypesMappers').value() ?? [];
+      const seenPrevCts: ContentTypesMapper[] = [];
+      const seenPrevUids = new Set<string>();
+      for (let i = 1; i < iteration; i++) {
+        const priorModel = getContentTypesMapperDb(projectId, i);
+        await priorModel.read();
+        const cts = priorModel.chain.get('ContentTypesMappers').value() ?? [];
+        for (const ct of cts) {
+          const uid = ct?.otherCmsUid;
+          if (uid && !seenPrevUids.has(uid)) {
+            seenPrevUids.add(uid);
+            seenPrevCts.push(ct);
+          }
+        }
+      }
 
-      const filtered = filterContentTypesByIteration(content_mapper, prevContentMapper, filter);
+      const filtered = filterContentTypesByIteration(content_mapper, seenPrevCts, filter);
       content_mapper.length = 0;
       content_mapper.push(...filtered);
 
@@ -1971,7 +1985,7 @@ const getExistingExtensions = async ({existingStackId, token_payload}: any) => {
 
 const updateEntryStatus = async (req: Request) => {
   const { projectId } = req?.params;
-  const { ids } = req?.body;
+  const { ids, locale } = req?.body;
   const validatedUids: string[] = Array.isArray(ids) ? ids : [];
   const srcFunc = "updateEntryMapping";
   if (isEmpty(validatedUids)) {
@@ -1998,14 +2012,19 @@ const updateEntryStatus = async (req: Request) => {
     const EntryMapperModel = getEntryMapperDb(projectId, iteration);
     await EntryMapperModel.read();
     const foundEntry: EntryMapper[] = [];
-    // Rows in entry_mapper are already per-(entry × source-locale), so each id uniquely
-    // identifies one locale variant; toggling isUpdate directly is correct.
+    // Rows in entry_mapper are per-(entry × source-locale); each id is unique per row.
+    // Also scope the toggle by source-locale as a safety net so a same-id collision
+    // (if it ever happens) can't flip a sibling locale's row and clobber the user's
+    // selection state on the other locale.
+    const sourceLocale = locale
+      ? getSourceLocaleForDestination(projectData ?? {}, locale)
+      : null;
     await EntryMapperModel.update((data: any) => {
       data?.entry_mapper?.forEach((entry: any) => {
-        if (validatedUids.includes(entry?.id)) {
-          entry.isUpdate = !entry.isUpdate;
-          foundEntry.push(entry);
-        }
+        if (!validatedUids.includes(entry?.id)) return;
+        if (sourceLocale && (entry?.language ?? '') !== sourceLocale) return;
+        entry.isUpdate = !entry.isUpdate;
+        foundEntry.push(entry);
       });
     });
 
--- a/api/src/services/contentMapper.service.ts
+++ b/api/src/services/contentMapper.service.ts
@@ -37,6 +37,7 @@ import { getSourceLocaleForDestination } from '../utils/locale-migration.utils.j
 import { loadPreviousAssetMetadata } from '../utils/asset-update.utils.js';
 import { flattenNestedUidMap } from '../utils/uid-mapper.utils.js';
 import { contentfulService } from './contentful.service.js';
+import { sanitizeStackId, assertResolvedPathUnderBase } from '../utils/sanitize-path.utils.js';
 
 
 const idCorrector = ({ id }: { id: string }) => {
@@ -2527,14 +2528,22 @@ const getAssetMapping = async (req: Request) => {
     // after an actual migration run, so it's absent (empty status) before that.
     let failedAssets: Record<string, any> = {};
     const destinationStackId = projectData?.destination_stack_id;
-    if (destinationStackId) {
-      const failedPath = path.join(MIGRATION_DATA_CONFIG.DATA, destinationStackId, MIGRATION_DATA_CONFIG.ASSETS_DIR_NAME, MIGRATION_DATA_CONFIG.ASSETS_FAILED_FILE);
-      if (fs.existsSync(failedPath)) {
-        try {
+    // destination_stack_id is DB-stored, but Snyk's taint tracker still traces it back to
+    // the HTTP projectId param via the lowdb lookup above — sanitize it the same way the
+    // rest of this codebase does (sanitizeStackId strips it to an allowlisted charset,
+    // assertResolvedPathUnderBase re-confirms the joined path can't escape the data dir)
+    // before it reaches a readFileSync sink.
+    const safeDestinationStackId = sanitizeStackId(destinationStackId);
+    if (safeDestinationStackId) {
+      const assetsBase = path.resolve(MIGRATION_DATA_CONFIG.DATA);
+      const failedPath = path.join(assetsBase, safeDestinationStackId, MIGRATION_DATA_CONFIG.ASSETS_DIR_NAME, MIGRATION_DATA_CONFIG.ASSETS_FAILED_FILE);
+      try {
+        assertResolvedPathUnderBase(assetsBase, failedPath);
+        if (fs.existsSync(failedPath)) {
           failedAssets = JSON.parse(fs.readFileSync(failedPath, 'utf-8')) || {};
-        } catch {
-          failedAssets = {};
         }
+      } catch {
+        failedAssets = {};
       }
     }
     const enrichedMapping = uidEnriched.map((item: any) => {
--- a/api/src/services/contentful.service.ts
+++ b/api/src/services/contentful.service.ts
@@ -809,7 +809,6 @@ const createAssets = async (packagePath: any, destination_stack_id: string, proj
 
       await Promise.all(tasks);
       await fs.promises.mkdir(assetsSave, { recursive: true });
-      const assetMasterFolderPath = path.join(assetsSave, ASSETS_FAILED_FILE);
 
       await writeOneFile(path.join(assetsSave, ASSETS_SCHEMA_FILE), assetData);
       // This code is intentionally commented out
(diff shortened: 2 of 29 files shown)


api/src/services/contentful/jsonRTE.ts
```diff
@@ -385,34 +385,61 @@ function parseHeading6(obj: any): any {
...
+//   - entry hyperlink (link ref) → type: 'reference', display-type: 'link', type: 'entry'
+//   - asset hyperlink (link ref) → type: 'reference', display-type: 'link', type: 'asset'
+// The previous implementation used non-standard types ('hyperlink', 'entry-hyperlink',
+// 'asset-hyperlink') that Contentstack's JSON RTE reader silently dropped, so URLs
+// vanished in the destination stack even though the surrounding text migrated.
+
 function parseEntryHyperlink(obj: any, lang?: LangType): any {
+  const targetSys = obj?.data?.target?.sys ?? {};
+  const entryUid = targetSys?.id ?? '';
+  const contentTypeUid = targetSys?.contentType?.sys?.id ?? '';
```
umesh-more-cstk: **blocker:** `target.sys.contentType` does not exist on Contentful export data. In a `contentful-export` JSON, a rich-text `entry-hyperlink` target is an unresolved Link — `{ sys: { type: 'Link', linkType: 'Entry', id: '...' } }` — with no `contentType`. So `contentTypeUid` is always `''`, and the emitted node is `{'entry-uid': '<src id>', 'content-type-uid': ''}`, which can't be resolved to a destination reference.

That's precisely why the sibling parsers don't try to read it: `parseBlockReference` (218-249) and `parseInlineReference` (251-280) look the CT uid up from the rte-references file, and `createRefrence` (`contentful.service.ts:1606`) writes it as snake_case (`name.replace(/([A-Z])/g,"_$1").toLowerCase()`) — so even a populated Contentful `contentType.sys.id` (`blogPost`) would be the wrong value (`blog_post`).

Suggest mirroring the existing pattern:

```ts
function parseEntryHyperlink(obj: any, lang?: LangType, destination_stack_id?: StackId): any {
  const targetId = obj?.data?.target?.sys?.id ?? '';
  const rteRefs: Record<string, any> =
    (destination_stack_id &&
      readFile(path.join(process.cwd(), DATA, destination_stack_id, RTE_REFERENCES_DIR_NAME, RTE_REFERENCES_FILE_NAME))) || {};
  const contentTypeUid =
    rteRefs?.[String(lang).toLowerCase()]?.[targetId]?._content_type_uid ?? '';
  // …then bail out (or fall back to a plain 'a' node) when contentTypeUid is empty,
  // rather than emitting a reference that can never resolve.
```

Note `parseEntryHyperlink` currently doesn't receive `destination_stack_id`; the `parsers` map already passes it, so widening the signature is enough. Same node also feeds the HTML path (`contentful.service.ts:471` `jsonToHtml`), so worth eyeballing an `html` RTE field with an entry link too.

api/src/services/contentful.service.ts
```diff
@@ -662,8 +662,28 @@ const saveAsset = async (
...
+      // assets whose CDN entry is ready, and `fields.file[locale].upload` (absolute fetch URL)
+      // for assets that were just added to the space but not yet processed. Real deltas hit
+      // the second shape on "newly added asset" exports; reading only `.url` silently drops
+      // those (see CMG-1106). Prefer `.url`, fall back to `.upload`, skip if neither exists.
+      const fileMeta = Object.values(assets?.fields?.file)[0] as { url?: string; upload?: string; contentType?: string; details?: { size?: string }; fileName?: string };
+      let fileUrl = '';
+      if (typeof fileMeta?.url === 'string' && fileMeta.url) {
+        fileUrl = fileMeta.url.startsWith('//') ? `https:${fileMeta.url}` : fileMeta.url;
+      } else if (typeof fileMeta?.upload === 'string' && fileMeta.upload) {
+        fileUrl = fileMeta.upload;
```
umesh-more-cstk, line 675: **blocker:** this fallback correctly lets an unprocessed asset (`.upload`, no `.url`) through, but nothing below it tolerates that shape, so the CMG-1106 case still fails — just later and more silently.

An asset that only has `.upload` has no `details` and often no `url`/`fileName` yet (Contentful only populates `details` after processing). Downstream, in this same function:

- `688-691`: `path.basename((Object.values(assets.fields.file)[0] as { fileName: string }).fileName)` → `TypeError` if `fileName` is absent.
- `725`: `` `${(… as {details:{size:string}})?.details.size}` `` → the `?.` is on the *file object*, not on `details`, so a missing `details` throws.
- `759`: same expression in the `retryCount === 1` failure branch, with no optional chaining at all.

Concrete failure path: `axios.get(fileUrl)` succeeds → building `assetData[...]` throws at 725 → inner `catch` → recursive retry (**second download of the same file**) → throws again at 725 → `catch` with `retryCount === 1` → 759 throws → escapes to the outer `catch (error) { console.error(error) }`. Net result: asset downloaded twice, dropped from the migration, and **no `failedJSON` entry**, so it never appears in the failed-assets report either.

Fix (both spots) plus a filename guard:

```ts
const fileSize = `${fileMeta?.details?.size ?? ''}`;
const rawFileName = typeof fileMeta?.fileName === 'string' && fileMeta.fileName
  ? fileMeta.fileName
  : `${assets.sys.id}`;
const fileName = path.basename(rawFileName);
```

and use `fileSize` / `fileMeta?.contentType ?? ''` at 718-725 and 755-759 instead of re-deriving from `Object.values(...)`.

api/src/services/migration.service.ts
```diff
@@ -1274,6 +1283,58 @@ const startMigration = async (req: Request): Promise<any> => {
...
+            ...entryByLocaleKeys,
+          ]),
+        );
+        await recordMigratedLocales(projectId, ranLocales);
+      } catch (err) {
+        await customLogger(projectId, destinationStackId, 'warn', `Failed to record migrated locales: ${(err as Error)?.message}`);
+      }
     }
     else{
       await customLogger(projectId, destinationStackId, 'warn', 'No config file generated for delta migration; skipping update CLI step.');
```
umesh-more-cstk: **question:** with this PR, locale recording happens in exactly two places: `runCli` (only when `iteration <= 1`) and the block above (only when `configFilePath` is truthy). This `else` branch — iteration ≥ 2 with no config file — now records **nothing**.

`removeEntriesFromDatabase` returns `null` when `!entryMapperItems?.length || !stackId` (entry-update.utils.ts:75-78) or when the entries dir is missing (110-113), and `assetUpdates` is empty in those runs, so `configFilePath` stays `null` — yet `runCli` above *did* bulk-import, including any brand-new locale. Because `getMigratedLocales` only backfills when `migrated_locales` is `undefined` (locale-migration.utils.ts:44-52), and iteration 1 already wrote the array, that locale stays permanently absent → `isFullMigrationForLocale` keeps returning `true` → `removeEntriesFromDatabase` keeps routing all its entries through the localize path on every later restart. That's the same failure class this PR is fixing, via the other trigger.

Suggest recording the union of `master_locale` + `entryByLocale` here too (the update CLI didn't run, so `extractLocalesFromUpdateConfig` isn't relevant), e.g. hoist the `entryByLocale` read + `recordMigratedLocales` into a small local helper and call it from both branches.

Minor, same block: `assertResolvedPathUnderBase` failing at 1309 silently degrades to `updateConfig = null`, which loses every locale that the update CLI just localized. A `customLogger('warn', …)` in that inner `catch` would make it debuggable.

api/src/services/contentMapper.service.ts
```diff
@@ -2396,17 +2405,30 @@ const getAssetMapping = async (req: Request) => {
...
+    // uid. The user selects which of those to update with the current file's newer
+    // version. Brand-new assets in this iteration have no prior uid; they upload
+    // automatically during the run and don't need a Map Entry row (nothing to
+    // select or update yet). Iteration 1 is untouched — everything is new then.
+    const displayMapping = iteration > 1
+      ? enrichedMapping.filter((item: any) => {
+          const uid = item?.contentstackAssetUid;
+          return uid != null && String(uid).trim() !== '';
+        })
+      : enrichedMapping;
```
umesh-more-cstk, line 2583: **question:** this is a user-visible behaviour change (assets with no resolved `contentstackAssetUid` disappear from the Assets tab on iteration ≥ 2) and I can't map it to any of the six linked tickets — CMG-1097 is only the empty-state layout. Was this intended to ship here?

Two things I'd like confirmed:
1. The filter keys off `contentstackAssetUid`, which for rows saved before an import is filled from `uidMapper[iteration].assets` / `[iteration-1].assets` (2387-2404). Any asset that *was* migrated but whose key is missing from those two maps (e.g. a CLI-deduped asset whose mapping never landed, or a project on iteration 4 where the merge chain broke) now vanishes from the tab entirely with no indication — instead of showing up unselectable. Worth a log line at least.
2. On a delta iteration whose source is all-new assets, the tab is now legitimately empty — which is exactly the state the new SCSS empty-state styles render. Is "empty Assets tab" the intended UX there, or should new assets still be listed read-only so the user can see them?

api/src/services/contentMapper.service.ts
```diff
@@ -1997,14 +2012,19 @@ const updateEntryStatus = async (req: Request) => {
...
+      ? getSourceLocaleForDestination(projectData ?? {}, locale)
+      : null;
     await EntryMapperModel.update((data: any) => {
       data?.entry_mapper?.forEach((entry: any) => {
-        if (validatedUids.includes(entry?.id)) {
-          entry.isUpdate = !entry.isUpdate;
-          foundEntry.push(entry);
-        }
+        if (!validatedUids.includes(entry?.id)) return;
+        if (sourceLocale && (entry?.language ?? '') !== sourceLocale) return;
```
umesh-more-cstk: **nit:** the guard is fine (and matches the read path — `getEntryMapping` filters rows with the identical `(row?.language ?? '') === sourceLocale` predicate at 2141-2146, so ids handed back by the UI will match), but the response below still says success even when this filter matches zero rows.

`foundEntry` is `EntryMapper[]`, so `if (foundEntry)` (≈2030) is always truthy → HTTP 200 → the UI sets `persistedRowIds` and toasts "Entries saved successfully" with nothing persisted. Reachable if the project's locale mapping changes between the list load and the save, or for a CMS whose `entry_mapper` rows carry an empty `language` (e.g. `migration-drupal/libs/extractEntries.js` uses `row?.langcode || ''`).

```ts
if (foundEntry.length) {
  return { status: HTTP_CODES?.OK, data: foundEntry };
}
```

so the existing `NOT_FOUND` branch actually becomes reachable. (`removeEntriesFromDatabase` handles the untagged-`language` case with a `rowByUid` fallback — worth considering the same fallback here for consistency.)

umesh-more-cstk (commented):
## Review summary

Six delta defects in one PR. I traced each fix through the real code paths (uid-mapper write order in `runCli`, `removeEntriesFromDatabase` config-key shape, Venus `InfiniteScrollTable` internals, `createRefrence` output shape) rather than reading the diff alone. Most of the reasoning is sound and several fixes are provably correct. Two changed code paths look broken in ways that will show up on real exports, and there are a few undisclosed behaviour changes worth a conscious decision before merge.

### Blocking
1. **`api/src/services/contentful/jsonRTE.ts:399`** — `parseEntryHyperlink` reads `content-type-uid` from `target.sys.contentType.sys.id`, which does not exist on a Contentful export *Link* (`{sys:{type:'Link',linkType:'Entry',id}}`). It will always be `''`, and even when present it would be the camelCase Contentful id, not the snake_case CT uid this repo generates. `parseBlockReference` (lines 218-249) already solves exactly this by reading `_content_type_uid` out of the rte-references file.
2. **`api/src/services/contentful.service.ts:675`** — the new `.upload` fallback lets *unprocessed* Contentful assets reach the download path, but the code immediately below still dereferences `fields.file[locale].fileName` (688-691) and `.details.size` (725, and again 759 with no optional chaining). Unprocessed assets have no `details` (that is why they have no `.url`), so the very case CMG-1106 targets throws, downloads twice, and is dropped with **no** `failedJSON` entry.

### Non-blocking (questions / nits)
- Locale bookkeeping now lives only inside `if (configFilePath)` in `migration.service.ts` while `runCli` only records for iteration ≤ 1 — iteration ≥ 2 runs where `removeEntriesFromDatabase` returns `null` record nothing at all (see inline).
- `getAssetMapping`'s new `displayMapping` filter is a user-visible behaviour change with no ticket in the description.
- `entryMapper.tsx`: entries are now fetched *only* from the locale effect, so a project whose locale map can't be resolved shows a permanently empty Map Entry table; and each content-type click now fires two identical requests.
- `MigrationLogViewer`: `'Migration Process Completed'` is written by `runCli` *before* the delta update CLI runs, so adding `'Entry Update Process Completed'` as an OR doesn't stop the early completion flip on iteration 2+.
- Reference resolution in `entry-update-script.cjs` is top-level-only (group / modular-block nesting still keeps source uids).
- Small doc/code mismatch in `flattenNestedUidMap`, a truthiness check in `updateEntryStatus`, and a `pickEntryTitle` edge case — all inline.

### Scope / process (out of diff, so noted here)
- The description claims **dep bumps for `fast-uri` / `brace-expansion` in root and each sub-project** plus a root `package.json` entry under Affected Areas. There is **no `package.json` or lockfile in this diff at all** (18 files, all source/test). Either the bumps were dropped or the description is stale — please reconcile, since the stated reason for keeping them here was to unblock the Snyk SLA check.
- Conversely, four changes in the diff have no listed ticket: the asset `.upload` fallback (`contentful.service.ts` + `extractAssets.js`, code comment references CMG-1106 which isn't in the ticket list), the `getAssetMapping` delta display filter, the `MigrationLogViewer` completion/purge rework, and the `migrationCompleted: false` reset in `pages/Migration/index.tsx`. Each is defensible, but they widen the blast radius beyond the six linked tickets.

### Verified good (worth calling out)
- `contentful.service.ts:439` `Array.isArray(entryId?.[id])` is a genuine fix — the old `entryId?.id` tested a literal `id` property, so the array branch was dead and already-array values got double-wrapped.
- `parseAssetHyperlink`'s attrs (`asset.uid` / `asset.filename` / `asset.content_type`) match exactly what `saveAsset` writes into the assets schema — unlike the pre-existing `parseBlockAsset`, which reads `asset.fileName` and gets `undefined`.
- `parseHyperlink` → `type:'a'` + `attrs.url` is the correct Contentstack JSON RTE anchor shape.
- Deferring locale recording until after `updateEntryCli` is right: `updateEntryCli.service.ts:240-249` really does swallow failures, so the old position could permanently skip a locale.
- `writeUidMapping` runs **before** `writePerLocaleEntryUidMapping` in `runCli` (it assigns `data = {assets, entry}` wholesale), so the new `entryByLocale` read in `migration.service.ts` isn't reading a clobbered key. Good ordering.
- `setTableRevision` is correctly inside the `seedSelection` branch, so pagination/search don't remount the table; and I confirmed in the bundled Venus source that `getSelectedRow` is called with `Object.keys(selectedRowIds)` (seeded from `initialSelectedRowIds`), so dropping `initialRowSelectedData` does **not** change the id set the app receives — only the unused second arg / `maxSelect` bookkeeping.
- `extractAssets.js`'s new `continue` is inside a `for…of`, so it's valid; and locale codes flowing into `migrated_locales` are consistently *destination* codes across `extractLocalesFromUpdateConfig`, `entryByLocale` and `master_locale` keys.

### Needs a human at a browser
The `index.scss` empty-state rules can't be verified statically. One thing to check: `.asset-mapper-table .Table__body` already carries `flex: 0 0 auto !important` (line ~858), so the new `display:flex; flex-direction:column` on the body won't let it grow into the forced `calc(100vh - 22rem)` table height — the illustration may be centred only because of `min-height: 300px` on `.Table__centerWrapper`. Please confirm at a couple of viewport heights / zoom levels.

api/src/utils/entry-update-script.cjs
```diff
@@ -4,6 +4,72 @@ const isAssetField = (value) =>
...
+    const oldFlat = entryMapping?.old?.flat?.[sourceUid];
+    if (oldFlat) return oldFlat;
+    return sourceUid;
+};
+
+/**
+ * Remaps the uid(s) inside a reference field value (single link object or
+ * array of link objects) to their Contentstack destination uids.
+ */
+const resolveReferenceField = (fieldName, entryUid, value, locale, entryMapping) => {
```
umesh-more-cstk, line 52: **question:** the resolver only walks *top-level* field values — `isReferenceValue(v)` / `isReferenceArray(v)`. References nested one level deeper aren't remapped, so on the delta/localize path they keep the source-CMS uid, which is the exact bug this fixes for flat fields.

Concretely, `processField` in `contentful.service.ts` returns nested structures for group / modular-block-ish fields (the `default` branch at 509-515 → `processArrayFields`, which pushes `entryId[id]` objects *inside* arrays of objects, and the `group` branch at 504). A payload like `{ heroBlock: { cta: { uid: '<src>', _content_type_uid: 'page' } } }` or `{ blocks: [{ card: { ref: {uid,_content_type_uid} } }] }` falls through `resolveReferenceField` untouched.

Is that a known limitation for this iteration? If so a `TODO` naming it would help; otherwise a small recursive walk (recurse into plain objects and into array items that aren't reference values) covers it without changing the flat behaviour.

api/src/utils/uid-mapper.utils.ts
```diff
@@ -5,6 +5,28 @@ import customLogger from "./custom-logger.utils";
...
+ * Kept in this module so both entry-mapping consumers (`contentMapper.service`
+ * and `entry-update.utils`) share one authoritative implementation.
+ */
+export const flattenNestedUidMap = (raw: Record<string, any> | undefined | null): Record<string, any> => {
+  const keys = Object?.keys(raw ?? {});
+  if (keys?.length === 0) return {};
+  const nested = keys?.every((k) => {
+    const v = (raw as Record<string, any>)[k];
+    return v != null && typeof v === 'object' && !Array.isArray(v);
+  });
```
umesh-more-cstk: **nit:** the doc and the code disagree. Line 11 says "If **ANY** top-level value is a non-array object, we treat the whole map as nested", but line 19 uses `.every(...)` — it's ALL, not ANY. (The behaviour is carried over unchanged from `contentMapper.service.ts`, so this is just the new comment being wrong.)

Worth noting the consequence, since this now feeds `__entryMapping__.flat`: a *mixed* map (some string values, some nested objects — which `mergeUidMaps` at 37-54 can produce when a prior iteration stored the flat shape and the current run the nested one) takes the pass-through branch, so `flat[someCtUid]` ends up being an **object**. `resolveReferenceUid` would then happily return that object as a uid if a source uid ever collided with a content-type uid. Either fix the comment to say "every", or make it robust:

```ts
const out: Record<string, any> = {};
for (const k of keys) {
  const v = raw[k];
  if (v != null && typeof v === 'object' && !Array.isArray(v)) Object.assign(out, v);
  else out[k] = v;
}
return out;
```

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -200,9 +213,11 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
       setContentTypeUid(data?.contentTypes?.[0]?.id ?? '');
       setOtherCmsUid(data?.contentTypes?.[0]?.otherCmsUid ?? '');
-      if (data?.contentTypes?.[0]?.id) {
-        fetchEntries(data?.contentTypes?.[0]?.id, searchVal ?? '', { seedSelection: true });
-      }
+      // Don't fetch entries here — the locale-effect at [selectedLocale.value] will fire
+      // once both contentTypeUid and selectedLocale are ready. Fetching now would run
+      // without a locale filter (selectedLocale is still null on initial mount), so the
+      // server would return all-locale rows and Venus's Table would snapshot that mixed
+      // selection state before the correct per-locale fetch could overwrite it.
```
umesh-more-cstk, line 238: **question:** removing the fetch here makes the entry table strictly dependent on `selectedLocale` ever becoming non-null. `selectedLocale` is only set in the locale effect (163-184) from `getProject`, and that effect just `console.error`s on failure. So if `getProject` fails, or the project has no `master_locale`/`locales` keys (legacy project, or `opts.length === 0` at 179), `selectedLocale` stays `null` → the effect at 193-198 never fires → **Map Entry renders an empty table forever**, with no spinner and no error.

A cheap safety net keeps the locale-race fix while avoiding the dead end:

```ts
} finally {
  setLocalesResolved(true);   // or: if (!opts.length) fetchEntries(ctId, '', { seedSelection: true })
}
```

i.e. once locale resolution has *settled* with no options, fall back to the unfiltered fetch (server-side `getEntryMapping` already "falls open when no locale is provided", contentMapper.service.ts:2141-2146).

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -176,13 +185,17 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
...
+  // types load before/after the locale mapping — whichever resolves last triggers the fetch,
+  // so we never call the API without a locale filter (which would return mixed-locale rows
+  // and clobber the Venus Table's initial selection snapshot).
   useEffect(() => {
     if (contentTypeUid && selectedLocale?.value) {
       fetchEntries(contentTypeUid, searchText || '', { seedSelection: true });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [selectedLocale?.value]);
+  }, [selectedLocale?.value, contentTypeUid]);
```
umesh-more-cstk, line 206: **nit:** adding `contentTypeUid` to the deps means every content-type selection now fires **two** identical requests: the click handler calls `fetchEntries(ct.id, …)` directly (`handleContentTypeClick`, ≈277) *and* `setContentTypeUid(ct.id)` re-triggers this effect. Same for the filter-reset path at ≈337. The new `fetchGenerationRef` guard makes it harmless (the loser is discarded), but it doubles server load on every click.

Since the effect is now the single source of truth for "contentTypeUid or locale changed → refetch", the two direct `fetchEntries(...)` calls in `handleContentTypeClick` / the filter handler can just be dropped.

ui/src/components/LogScreen/MigrationLogViewer.tsx
```diff
@@ -190,12 +194,27 @@ const MigrationLogViewer = ({ serverPath }: LogsType) => {
...
+    // Full/master import ends with "Migration Process Completed"; the delta update path ends
+    // with "Entry Update Process Completed" — accept either as the terminal message.
+    const TERMINAL_MESSAGES = new Set([
+      'Migration Process Completed',
+      'Entry Update Process Completed'
+    ]);
+    const hasTerminalMessage = logs?.some(
+      (log) => log?.message && TERMINAL_MESSAGES.has(log.message)
+    );
+    if (migrationStarted && hasTerminalMessage) {
```
umesh-more-cstk: **question:** accepting either message as terminal means `'Migration Process Completed'` still wins on the delta path, so I don't think this changes the early-completion behaviour on iteration 2+.

On iteration ≥ 2 the order is: `runCli` writes `'Migration Process Completed'` (runCli.service.ts:258-273) → *then* `migration.service.ts` runs `enrichConfig*` + `utilsUpdateCli.updateEntryCli`, which is what finally writes `'Entry Update Process Completed'` (updateEntryCli.service.ts:233-239). Since this is an OR over the whole array, the modal + `migrationStarted:false / migrationCompleted:true` dispatch fires as soon as the bulk import ends — while the localize/update pass (and its `'Failed to update entries'` error line) is still streaming, now hidden behind the completion view.

If the intent is "delta ends only when the update CLI ends", the terminal message needs to be iteration-aware, e.g. require `'Entry Update Process Completed'` when `iteration > 1` and `'Migration Process Completed'` otherwise.

Separately, on the purge above (134): FYI the server never replays — `server.ts:117-142` streams from a monotonic `fileOffset`, so previously-emitted lines are not re-sent on reconnect. The purge is still harmless in the normal flow, but the stated justification ("replayed … on socket reconnect / late buffered emit") only actually holds after an **api restart** (which resets `fileOffset` to 0 and re-emits the whole file to every client).

upload-api/migration-contentful/libs/extractEntries.js
```diff
@@ -9,6 +9,33 @@ const { readFile } = require('../utils/helper');
...
+  }
+  let hasAnyLocaleContent = false;
+  for (const val of Object.values(fields)) {
+    if (val == null || typeof val !== 'object') continue;
+    if (!(locale in val)) continue;
+    hasAnyLocaleContent = true;
+    const localized = val[locale];
+    if (typeof localized === 'string' && localized.trim()) return localized;
+  }
+  return hasAnyLocaleContent ? entry?.sys?.id : null;
```
umesh-more-cstk, line 40: **nit:** `hasAnyLocaleContent` is set from key *presence* (`locale in val`), not from the value being meaningful, so it doesn't match the JSDoc ("Returns null when the entry has no content at this locale at all").

A field serialised as `{ "en-US": "", "de-DE": "x" }` — or `{ "en-US": null }`, or a boolean `false` — flips the flag, so an entry that isn't really localised to `en-US` now emits a Map Entry row titled with its raw `sys.id`. Previously (title/name only) it was skipped. On a multi-locale export that inflates the per-locale row counts and shows users opaque `sys.id` rows.

```js
const localized = val[locale];
if (localized === null || localized === undefined || localized === '') continue;
hasAnyLocaleContent = true;
if (typeof localized === 'string' && localized.trim()) return localized;
```

Also worth noting the step-3 fallback returns the *first* non-empty string field in `Object.values(fields)` insertion order, which can surface a slug/URL rather than a human title — deterministic but arbitrary; the `displayField` path (the actual CMG-1102 fix) covers the cases that matter, so this is only about the fallback.

api/src/services/migration.service.ts
```diff
@@ -1274,12 +1283,95 @@ const startMigration = async (req: Request): Promise<any> => {
...
+      await recordDeltaMigratedLocales(
+        projectId,
+        safePid,
+        iteration,
+        project,
+        destinationStackId,
+        configFilePath,
+      );
     }
     else{
```
umesh-more-cstk, line 1301: **Locale recording is skipped entirely when no update config exists.**

This block runs only inside `if (configFilePath)`. But `removeEntriesFromDatabase` returns `null` when the entry mapper is empty / stackId is missing or the entries directory doesn't exist (`entry-update.utils.ts:75-78, 110-113`), and `ensureUpdateConfigFile` only kicks in when there are asset updates. In that case the run falls into the `else` branch and records nothing — and since `runCli.service.ts` now deliberately skips recording for iteration > 1, `migrated_locales` never gets updated for that iteration.

So a delta iteration whose bulk import created only brand-new entries (the exact case "source 3" in the comment above exists for) would never mark its locales as migrated, and those locales get routed through the localize path on every restart — the same failure mode this PR fixes, just via a different trigger.

Suggestion: run the master-locale + `entryByLocale` union recording unconditionally after `runCli` resolves, and keep only the `extractLocalesFromUpdateConfig` contribution behind `configFilePath`.

---

api/src/services/contentMapper.service.ts
```diff
@@ -1997,18 +2012,23 @@ const updateEntryStatus = async (req: Request) => {
...
```
umesh-more-cstk: **Legacy rows without a `language` tag can never be toggled once the UI sends a locale.**

For rows created before `language` tagging existed, `(entry?.language ?? '') !== sourceLocale` is always true, so the toggle silently no-ops (and the request still returns 200 with an empty `foundEntry`). `removeEntriesFromDatabase` handles this same situation with a `rowByUid` fallback — but this path has no equivalent.

Since this guard is framed as a safety net, consider only enforcing it when the row actually carries a language:

```ts
if (sourceLocale && entry?.language && entry.language !== sourceLocale) return;
```

---

api/src/utils/uid-mapper.utils.ts
```diff
@@ -5,6 +5,31 @@ import customLogger from "./custom-logger.utils";
 import fs from "fs";
 import projectModelLowdb from "../models/project-lowdb";
 
+/**
+ * Normalises a uid map that may be either flat `{ sourceUid: destUid }` or nested
+ * per-content-type `{ [ctUid]: { sourceUid: destUid } }` into a single flat map.
+ * Checked per-key rather than all-or-nothing, so a MIXED map (some flat string
```
umesh-more-cstk, line 11: Nit: the comment says "If ANY top-level value is a non-array object, we treat the whole map as nested", but the implementation (correctly, matching the previous inline version in `contentMapper.service.ts`) uses `.every(...)` — the map is treated as nested only when **all** values are non-array objects. Fix the comment to say "every".

---

ui/src/components/LogScreen/MigrationLogViewer.tsx
```diff
@@ -190,12 +196,31 @@ const MigrationLogViewer = ({ serverPath }: LogsType) => {
...
-    logs?.forEach((log) => {
+    // Look for a terminal message anywhere in `logs`. Scanning the whole array is safe here
+    // because the effect above purges `logs` back to the placeholder when migrationStarted
+    // flips false→true — anything present is from the current run, not a replay.
+    // We can't inspect only the last entry: after the CLI emits "Migration Process Completed"
+    // the backend still writes uid-mapper / "No config file generated" lines, burying the
+    // terminal message mid-array on the delta path.
+    const migrationStarted = newMigrationData?.migration_execution?.migrationStarted;
+    // On iteration 1 there's only a bulk import — "Migration Process Completed" IS terminal.
+    // On iteration 2+ (delta) that message only marks the bulk-import phase; the run isn't
```
umesh-more-cstk, line 207: Note (pre-existing, no action required for this PR): on the delta path `runCli` still writes `"Migration Process Completed"` to the log **before** `updateEntryCli` runs (`runCli.service.ts:260`), so this detector opens the completion modal while entry updates are still in flight. That behavior predates this PR, but since this is a rewrite of exactly this detector it's worth a follow-up ticket if not intended — e.g. suppressing the bulk-import terminal message on iteration > 1 and relying on `"Entry Update Process Completed"` alone.

---

api/src/services/contentful/jsonRTE.ts
```diff
@@ -385,34 +385,76 @@ function parseHeading6(obj: any): any {
...
+  // Prefer the anchor text from `obj.content` (Contentful nests the link label
+  // as a text child); fall back to a stale `target.title` if present.
+  const text = obj?.content?.[0]?.value ?? obj?.data?.target?.title ?? '';
+
+  // A Contentful export's entry-hyperlink target is an unresolved Link — it never
+  // carries `sys.contentType`. The destination content-type uid has to come from
+  // the rte-references file (the same source parseBlockReference/parseInlineReference
+  // use), keyed by locale then by target entry id.
+  const rteRefs: { [key: string]: any } | undefined =
+    destination_stack_id && readFile(path.join(process.cwd(), DATA, destination_stack_id, RTE_REFERENCES_DIR_NAME, RTE_REFERENCES_FILE_NAME));
```
umesh-more-cstk, line 413: Note: `entry-uid` carries the Contentful source id here, relying on the CLI import's uid-replacement pass to remap it — consistent with how reference fields work on the bulk path. But RTE reference nodes inside entries that go through the **update-script** path won't get remapped (`resolveReferenceField` only handles the `{uid, _content_type_uid}` shape, not RTE node trees). Before this PR those links were dropped entirely, so this is still a strict improvement — just flagging the residual gap for a follow-up.

---

umesh-more-cstk (commented):
Reviewed all six fixes against the surrounding delta code paths and ran the API test suite locally (79 files, 725 tests — all passing). Overall a solid PR: the fixes are well-scoped, the intent comments are genuinely useful, and the new tests cover the resolver precedence and locale-extraction logic well.

**Verified during review**

- The `tableRevision` remount only bumps on `seedSelection` fetches, so pagination/search don't reset the entry table.
- The UI sends `locale` in the `updateEntryStatus` body (`entryMapper.tsx:439`), so the server-side scoping is actually exercised.
- `__locale` (destination dir name) and `entryByLocale` keys agree, so the per-locale resolver lookups line up.
- The union-of-prior-iterations CT fix is safe: `filterContentTypesByIteration` only consumes `otherCmsUid` from the prior list, so keeping the first-seen CT object per uid is harmless.
- The `Array.isArray(entryId?.[id])` typo fix in `contentful.service.ts` is correct.

**Findings** (details in the line comments)

1. **Medium** — locale recording for iteration ≥ 2 is skipped entirely when no update config exists (`migration.service.ts`), which can reproduce the "locale permanently skipped" bug this PR fixes via a different trigger.
2. **Minor** — `updateEntryStatus`'s locale guard silently no-ops on legacy rows without a `language` tag.
3. **Nit** — `flattenNestedUidMap` doc comment says "ANY" where the code does `every`.
4. Two informational notes on pre-existing behavior (early completion modal on delta runs; RTE `entry-uid` not remapped on the update-script path).

**Description mismatch**: the PR description claims dep bumps (`fast-uri`, `brace-expansion`) across root and sub-project manifests, but the diff contains no `package.json` changes — Snyk reports "No manifest changes detected in 8 projects". Presumably those landed on `dev` separately; please update the description so release notes aren't misled.

---

ui/src/components/LogScreen/MigrationLogViewer.tsx
```diff
@@ -190,12 +196,31 @@ const MigrationLogViewer = ({ serverPath }: LogsType) => {
...
+    // actually done until the update/localize CLI finishes and writes
+    // "Entry Update Process Completed" afterward. Treating either as terminal on a delta run
+    // would fire the completion modal early, hiding the still-streaming localize pass (and any
+    // "Failed to update entries" error in it) behind the completion view.
+    const isDeltaIteration = (newMigrationData?.iteration ?? 1) > 1;
+    const requiredTerminalMessage = isDeltaIteration
+      ? 'Entry Update Process Completed'
```
umesh-more-cstk, line 214: **blocker:** On a delta iteration this makes `'Entry Update Process Completed'` the *only* accepted terminal signal, but there are two reachable paths on iteration ≥ 2 where that string is never written — and `MigrationLogViewer.tsx:232` is the only place in the whole UI that ever sets `migrationCompleted: true` (verified: `git grep 'migrationCompleted: true' ui/src` returns exactly one hit). So when neither path fires, the run finishes on the server and the UI never leaves the spinner.

Path 1 — update CLI skipped. In `api/src/services/migration.service.ts` the `else` branch (no `configFilePath`, and `assetUpdates.length === 0`, so `ensureUpdateConfigFile` isn't called either) logs only `'No config file generated for delta migration; skipping update CLI step.'` and never invokes `utilsUpdateCli.updateEntryCli`. That is the normal outcome for a delta iteration where the user selected no entries and no assets to update — e.g. an iteration that only adds brand-new content types.

Path 2 — update CLI failed. `api/src/services/updateEntryCli.service.ts:235` writes `'Entry Update Process Completed'` only on the success path; the `catch` writes `'Failed to update entries for stack: …'` and swallows the error. This PR's own comments in `migration.service.ts` and `runCli.service.ts` cite that swallow, so it's a known-live path.

Failure scenario: run iteration 2 with nothing selected on Map Entry and no asset updates. Bulk import succeeds, `'Migration Process Completed'` is emitted, the backend logs the "No config file generated" warning and returns. `isDeltaIteration` is true, `hasTerminalMessage` stays false forever → `setIsModalOpen(true)` never runs, `migrationStarted` stays `true`, `stepValue` never flips to `'Restart Migration'`. The user sits on Execution Logs indefinitely after a fully successful migration, with no way to reach the completion view or restart.

Suggested fix — give the delta path exactly one guaranteed terminal signal rather than inferring it from a phase-specific message. Cleanest is server-side: in `migration.service.ts`, emit `'Entry Update Process Completed'` (or a new `'Delta Migration Process Completed'`) after the whole delta branch — including the `else` branch and after `updateEntryCli` returns regardless of whether it threw. If it has to stay client-side, accept a set instead of a single literal:

```ts
const terminalMessages = isDeltaIteration
  ? [
      'Entry Update Process Completed',
      'Failed to update entries',            // updateEntryCli catch path
      'No config file generated for delta migration; skipping update CLI step.',
    ]
  : ['Migration Process Completed'];
const hasTerminalMessage = logs?.some((log) =>
  terminalMessages.some((m) => log?.message?.startsWith(m))
);
```
(the `Failed to update entries` line is `Failed to update entries for stack: ${stack_api_key}`, hence the prefix match).

---

api/src/services/contentMapper.service.ts
```diff
@@ -1997,18 +2012,23 @@ const updateEntryStatus = async (req: Request) => {
...
+        entry.isUpdate = !entry.isUpdate;
+        foundEntry.push(entry);
       });
     });
 
-    if (foundEntry) {
+    if (foundEntry.length) {
```
umesh-more-cstk, line 2046: **question:** Tightening `if (foundEntry)` → `if (foundEntry.length)` is correct in itself (the old check was always truthy — an array literal). But it turns a previously-silent no-op into a user-visible failure on a path this PR's own CMG-1104 fix walks through, and I don't think `updateEntryStatus` covers that path.

The read side, `getEntryMapping`, falls back to iteration `N-1`'s `EntryMapperModel` when the current iteration has no rows yet (`contentMapper.service.ts` ~2119-2131, "Fallback: If no entry mappings found in current iteration…"). `updateEntryStatus` has no such fallback — it only opens `getEntryMapperDb(projectId, iteration)`.

Failure scenario: iteration 2, project restarted but entry-mapper rows for iteration 2 not written yet. Map Entry renders rows sourced from iteration 1 via that fallback. User ticks a few and hits Save → `updateEntryStatus` scans iteration 2's empty `entry_mapper`, matches nothing, `foundEntry.length === 0` → 404. `ui/src/components/ContentMapper/entryMapper.tsx` `handleSaveContentType` only treats `status === 200` as success, so the user gets a red "Failed to save entries" toast and `persistedRowIds` is not updated — the selection is lost on the next fetch. Before this change the same call returned 200 with `data: []` and showed "Entries saved successfully" (also not persisting, but silently).

So this is strictly more honest than before, but it surfaces a real gap rather than closing it. Suggested fix: mirror the read path's fallback — if the current iteration's update matched nothing and `iteration > 1`, retry the toggle against `getEntryMapperDb(projectId, iteration - 1)` before returning 404. Alternatively, if the two are meant to be able to diverge, have `getEntryMapping` return the iteration each row came from so the client can pass it back.

Separately, worth confirming: `EntryMapperModel.update()` is still called (and writes the file) even when nothing matched. Harmless, but a no-op write per rejected save.

---

api/src/services/contentMapper.service.ts
```diff
@@ -2396,17 +2405,30 @@ const getAssetMapping = async (req: Request) => {
...
     });
 
-    if (!isEmpty(enrichedMapping)) {
+    // Delta migration intent: on iteration 2+ the Assets tab lists ONLY assets that
+    // were already migrated in a prior iteration — i.e. those with a Contentstack
+    // uid. The user selects which of those to update with the current file's newer
+    // version. Brand-new assets in this iteration have no prior uid; they upload
+    // automatically during the run and don't need a Map Entry row (nothing to
+    // select or update yet). Iteration 1 is untouched — everything is new then.
+    const displayMapping = iteration > 1
```
umesh-more-cstk: **question:** This is a substantive behaviour change to the Assets tab that no listed ticket covers — CMG-1097 is "Assets screen empty-state layout broken", i.e. CSS. Flagging it as scope as much as correctness: it belongs in the PR description (and arguably its own ticket) so QA knows to re-test asset selection on iteration 2+, not just the empty-state illustration.

On correctness, one interaction worth checking. `contentstackAssetUid` is either already on the row or backfilled just above from `uidMapperCurrent.data.assets` / `uidMapperPrev.data.assets`. Both of those reads can legitimately come back empty — `getUidMapperDb` creates the file with `defaultData = { entry: {}, assets: {}, entryByLocale: {} }`, so a fresh iteration directory yields `assets: {}` until the CLI import has run and `writeUidMapping` has written it.

Failure scenario: iteration 2, user reaches Map Entry → Assets tab before the import for this iteration has written its uid-mapper. The fallback above loads iteration 1's `asset_mapper` rows, but if those rows were persisted without `contentstackAssetUid` and iteration 1's uid-mapper `assets` map doesn't resolve them, `displayMapping` filters to `[]` and the tab renders empty with `count: 0` — indistinguishable from "no assets to update". Previously the rows were at least listed. Given the empty state is exactly what CMG-1097 is about, an all-filtered tab could easily be mistaken for the fix working.

Suggested fix: only apply the filter once you know the mapping data is actually loaded — e.g. skip it when both uid-mapper `assets` maps are empty AND no row carries a `contentstackAssetUid`, so an unresolved state shows rows rather than silently showing none. Or surface the distinction to the UI (`count: 0` + a `filtered: true` flag) so the empty state can say "no previously-migrated assets in this iteration" instead of the generic message.

---

api/src/services/contentful/jsonRTE.ts
```diff
@@ -385,34 +385,76 @@ function parseHeading6(obj: any): any {
...
-function parseEntryHyperlink(obj: any, lang?: LangType): any {
+// Contentstack JSON RTE uses:
+//   - plain hyperlink            → type: 'a',        attrs.url
+//   - entry hyperlink (link ref) → type: 'reference', display-type: 'link', type: 'entry'
+//   - asset hyperlink (link ref) → type: 'reference', display-type: 'link', type: 'asset'
+// The previous implementation used non-standard types ('hyperlink', 'entry-hyperlink',
+// 'asset-hyperlink') that Contentstack's JSON RTE reader silently dropped, so URLs
+// vanished in the destination stack even though the surrounding text migrated.
+
+function parseEntryHyperlink(obj: any, lang?: LangType, destination_stack_id?: StackId): any {
```
umesh-more-cstk, line 402: **question:** The new `destination_stack_id` parameter is threaded correctly through most container parsers, but not all of them — so CMG-1103 stays broken for hyperlinks nested in certain blocks.

`parseBlockquote` (line ~318) and `parseHeading1`…`parseHeading6` (lines ~328-378) all recurse with `parsers.get(e.nodeType)?.(e)` — no `lang`, no `destination_stack_id`. Every other container (`parseDocument`, `parseParagraph`, `parseLI`, the table parsers, …) passes all three. Since hyperlinks are inline nodes, a link inside a heading or blockquote is a realistic authoring shape.

Failure scenario: an ASIC-style entry with an entry-hyperlink inside an H2. `parseEntryHyperlink` is invoked with `destination_stack_id === undefined`, so `rteRefs` is `undefined`, `contentTypeUid` is `undefined`, and it takes the new `!targetId || !contentTypeUid` branch → returns `{ text }`. The label survives but the link is gone in the destination, which is the exact symptom CMG-1103 is meant to fix. Asset-hyperlinks in the same position are worse: `parseAssetHyperlink` returns `null` and the caller's `.filter(Boolean)` drops the node entirely, so the anchor text disappears too.

To be fair, this isn't a regression — before this PR `parseAssetHyperlink` would have thrown `TypeError: Cannot read properties of undefined` on `assetId[...]` in that same position, so the new null-guard is an improvement. But the fix is incomplete for these containers.

Suggested fix: give the seven parsers the same signature as their siblings and forward both args:

```ts
function parseBlockquote(obj: any, lang?: LangType, destination_stack_id?: StackId): any {
  const children = obj.content
    .map((e: any) => parsers.get(e.nodeType)?.(e, lang, destination_stack_id))
    .filter(Boolean);
  …
}
```
Same for `parseHeading1`–`parseHeading6`. Worth adding a fixture with a link inside a heading to the CMG-1103 test steps.

---

api/src/utils/entry-update-script.cjs
```diff
@@ -4,6 +4,98 @@ const isAssetField = (value) =>
...
+ * Recursively walks a field value and resolves any reference shape found at any
+ * depth — group and modular-block fields nest references one or more levels deep
+ * (see processField's 'group' branch and processArrayFields in contentful.service.ts),
+ * so a shallow top-level-only check misses them and they keep their source-CMS uid on
+ * the delta/localize path. Asset field objects are left untouched (they need
+ * resolveAssetField's 3-way stack comparison, not a uid remap) so this only ever
+ * rewrites reference shapes, nothing else.
+ */
+const resolveReferencesDeep = (fieldName, entryUid, value, locale, entryMapping) => {
+    if (isReferenceValue(value) || isReferenceArray(value)) {
```
umesh-more-cstk: **question:** The `isReferenceArray(value)` arm short-circuits the deep walk for the whole array. Because `isReferenceArray` was deliberately loosened to `.some()`, a *single* reference-shaped element is enough to route the entire array into `resolveReferenceField`, which maps element-wise and — per its own comment — passes non-reference items through **untouched**, never recursing into them.

So for an array where some elements are bare references and others are objects containing nested references, the nested ones keep their source-CMS uid. That's precisely the case `resolveReferencesDeep` was added to catch, per the docblock above ("group and modular-block fields nest references one or more levels deep").

Failure scenario: a modular-block field whose array holds one block object with a reference inside `{ heroBlock: { author: { uid, _content_type_uid } } }` alongside one top-level `{ uid, _content_type_uid }` element. `isReferenceArray` → true (`.some` matches the second), `resolveReferenceField` remaps the second and returns the first byte-for-byte, so `heroBlock.author.uid` is written to the destination as the Contentful source id and the reference dangles.

I couldn't construct this from the current Contentful producer — `processArrayFields` emits either all-reference arrays or all-block arrays, and the `[undefined]` case from the single-reference path in `contentful.service.ts` is a non-object, so in practice this looks unreachable today. Raising it as a latent trap rather than a live bug: the `.some()` loosening and the deep walk are individually right but compose into a hole, and a future producer (or another CMS's converter) emitting a mixed array would hit it silently.

Suggested fix — handle arrays uniformly and reserve `resolveReferenceField` for the single-value case, which removes the need for `isReferenceArray` here entirely:

```js
const resolveReferencesDeep = (fieldName, entryUid, value, locale, entryMapping) => {
    if (isReferenceValue(value)) {
        return resolveReferenceField(fieldName, entryUid, value, locale, entryMapping);
    }
    if (Array.isArray(value)) {
        return value.map((item) => resolveReferencesDeep(fieldName, entryUid, item, locale, entryMapping));
    }
    …
};
```
Every element then gets remapped if it's a reference and deep-walked if it's a container. Worth a unit test with a mixed array to pin the behaviour either way.

---

ui/src/components/ContentMapper/entryMapper.tsx
```diff
@@ -642,7 +687,12 @@ const EntryMapper = ({ handleStepChange }: entryMapperProps) => {
                 )}
                 <div className="entry-mapper-table">
                   <InfiniteScrollTable
-                  key={contentTypeUid || 'entry-mapper-table'}
+                  // `tableRevision` bumps AFTER a seedSelection fetch commits rowIds — so
+                  // the Table remounts with the correct initial selection already in state.
+                  // Keying on selectedLocale alone would remount too early (rowIds still
+                  // holding the previous locale's data, since the new fetch is still
+                  // in flight).
+                  key={`${contentTypeUid || 'entry-mapper-table'}::${tableRevision}`}
```
umesh-more-cstk, line 691: **nit:** Dropping the `initialRowSelectedData` prop from `InfiniteScrollTable` in this hunk leaves the backing state write-only. `initialRowSelectedData` is declared at line 113 and written at lines 255 and 398, but after this change nothing reads it — `ui/tsconfig.json` has no `noUnusedLocals`, so the build won't catch it.

Not a correctness bug, just dead state: `setInitialRowSelectedData(selectableInitialRows(validTableData))` still runs inside `fetchEntries`, costing a `selectableInitialRows` pass plus an extra render per fetch, on top of the `setTableRevision` remount happening a few lines later.

Suggested fix: delete the `useState` at line 113 and both setter calls. Worth confirming first that Venus's Table genuinely doesn't need the row-*data* form — `initialSelectedRowIds` alone carries ids, so if any select-all / header-checkbox behaviour was driven off `initialRowSelectedData` it would now be silently inert rather than failing loudly.

---

umesh-more-cstk (commented):
## Review — 1 blocker, 4 questions, 1 nit

Reviewed at head `4fb87ff`, base `dev` (`85b1049`) — 18 files, +1002/−163. Six delta-migration defects grouped into one PR. The core mechanisms hold up well: the union-of-prior-iterations fix for CMG-1095 is right, the locale plumbing is consistent (I traced `entryByLocale` → destination locale dir names in `writePerLocaleEntryUidMapping`, `__locale`/`::` suffix → `localeCode` in `removeEntriesFromDatabase`, and `migrated_locales` → `isFullMigrationForLocale` — all destination codes, no source/destination mismatch), the `updateEntryStatus` locale scoping matches `getEntryMapping`'s read-side filter exactly, and moving `recordMigratedLocales` after `updateEntryCli` resolves is the correct call. `flattenNestedUidMap` moving to `uid-mapper.utils.ts` with the per-key (rather than all-or-nothing) check is a genuine improvement over the version it replaces. Test coverage on the new helpers is good.

One blocker, on the UI completion path. The rest are questions where I want the author's read before calling them bugs.

### Blocker
- **`ui/src/components/LogScreen/MigrationLogViewer.tsx:214`** — on iteration ≥ 2 the completion detector requires the literal `'Entry Update Process Completed'`, which only `updateEntryCli`'s success path writes. Two reachable paths never write it (update CLI skipped entirely when there's no config file and no asset updates; update CLI's `catch`, which swallows the error). `MigrationLogViewer.tsx:232` is the only place in the UI that ever sets `migrationCompleted: true`, so in both cases the user is stranded on the Execution Logs spinner after a fully successful run. Details and two suggested fixes inline.

### Questions
- **`api/src/services/contentMapper.service.ts:2031`** — `if (foundEntry.length)` correctly replaces an always-truthy check, but `updateEntryStatus` lacks the previous-iteration fallback that `getEntryMapping` has, so rows rendered from iteration N−1 now save with a hard 404 + "Failed to save entries" toast instead of the old silent 200.
- **`api/src/services/contentMapper.service.ts:2414`** — the new iteration-2+ Assets filter can empty the tab when the uid-mapper `assets` map hasn't been written yet, which is indistinguishable from "nothing to update" — awkward given CMG-1097 is about that very empty state.
- **`api/src/services/contentful/jsonRTE.ts:396`** — `parseBlockquote` and `parseHeading1`–`parseHeading6` don't forward `lang`/`destination_stack_id`, so entry/asset hyperlinks nested in a heading or blockquote still lose their URL. CMG-1103 is only partly fixed.
- **`api/src/utils/entry-update-script.cjs:83`** — `.some()`-based `isReferenceArray` short-circuits the deep walk for mixed arrays, so nested references inside sibling block objects would be skipped. Looks unreachable with today's Contentful producer; raising it as a latent trap.

### Nit
- **`ui/src/components/ContentMapper/entryMapper.tsx:695`** — removing the `initialRowSelectedData` prop leaves the backing state write-only (declared L113, set L255/L398, never read).

### Scope

Flagging two things about the diff-vs-description fit — no action needed on the code itself, but the description should match what actually ships.

1. **The dependency bumps described aren't in this PR.** The description says "Dep bumps to clear Snyk SLA (`fast-uri`, `brace-expansion`) in root and each sub-project package.json", and Affected Areas checks "Other: root `package.json` (dep override bumps)". No `package.json` or lockfile appears in the diff against `dev`. Either the bumps were dropped during a rebase or the description is stale — worth confirming, because the Reviewer Notes cite unblocking the Snyk SLA check as the reason for keeping them in this PR, and that reason no longer applies. (For what it's worth, splitting them out would have been the cleaner call anyway, as the notes acknowledge.)

2. **Several substantive changes aren't attributable to any of the six linked tickets.** All within the checked Affected Areas, so the checkboxes are accurate, but "What changed?" doesn't mention them and QA won't know to test them:
   - `contentful.service.ts` `saveAsset` — `file.upload` fallback and defensive `fileName`/`details.size`/`contentType` handling. The code comment cites **CMG-1106**, which isn't in the ticket list. Good change (it removes a real `https:undefined` axios call and a `path.basename(undefined)` throw), just untracked.
   - `upload-api/migration-contentful/libs/extractAssets.js` — same `.upload` fallback, plus assets with no downloadable source are now **skipped entirely** rather than emitted with `assetPath: ''`. Behaviour change to Map Entry's asset list.
   - `contentMapper.service.ts` `getAssetMapping` — the iteration-2+ filter (see question above).
   - `MigrationLogViewer.tsx` — log purge on run start, single-pass terminal detection, and moving the "already migrated" placeholder out of the `.map` (that last one is a real fix — it previously rendered once per log line).
   - `ui/src/pages/Migration/index.tsx` — explicit `migrationCompleted: false` on start.
   - `contentful.service.ts:439` — `Array?.isArray(entryId?.id)` → `Array.isArray(entryId?.[id])`. The old expression tested a literal `id` key and was always false, so this actually activates a previously-dead branch and stops double-wrapping already-array reference values. Correct, but a real behaviour change hidden in a one-line diff. Note the `else` still returns `[entryId?.[id]]`, which yields `[undefined]` for an unmapped reference — the same `[undefined]` that `entry-update-script.cjs`'s new `isReferenceArray` comment works around. Worth fixing at the source rather than tolerating downstream.

Adding a line per item to the description (and a ticket for CMG-1106) would keep the delta-migration audit trail intact.

---

umesh-more-cstk (approved):

umesh-more-cstk (approved):

vikrantraut-cstk (approved):

PR #1137, opened 2026-08-06
Feat/skill

umesh-more-cstk:
## 🔗 Jira Ticket

> Replace with your ticket link — required before requesting review.

[MIGRATION-XXXX](https://contentstack.atlassian.net/browse/MIGRATION-XXXX)

---

## 📋 PR Type

- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

-

### Why?

---

## 🧩 Affected Areas

- [ ] `api` — Node.js backend
- [ ] `ui` — React frontend
- [ ] Other: 

---

## 🧪 How to Test

1.
2.
3.

**Expected result:**

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
|        |       |

---

-

---

## ✅ Author Checklist

> Complete this before moving the PR out of Draft.

- [ ] Branch follows naming convention: `feature/`, `bugfix/`, or `hotfix/` + 5–30 lowercase chars
- [ ] Jira ticket linked above
- [ ] Self-reviewed the diff — no debug logs, commented-out code, or TODOs left in
- [ ] `.env` / `example.env` updated if new environment variables were added
- [ ] No sensitive credentials or secrets committed

---

## 👀 Reviewer Notes

---


vikrantraut-cstk (approved):

upload-api/src/config/index.ts
```diff
@@ -29,7 +29,7 @@ export default {
     dropdown: { optionLimit: 100 }
   },
   // CMS type configuration
-  cmsType: process.env.CMS_TYPE || 'cmsType',
+  cmsType: process.env.CMS_TYPE || 'sanity',
```
umesh-more-cstk, line 32: **blocker:** Default `cmsType` is now `'sanity'`, but there is no Sanity connector in this repo yet.

`src/validators/index.ts` switches on `` `${type}-${extension}` `` and only handles `sitecore-zip`, `contentful-json`, `wordpress-xml`, `aem-folder`, `drupal-sql` — everything else hits `default: return false`.

Failure scenario: a deploy that doesn't set `CMS_TYPE` (the whole point of a default) uploads any file. `handleFileProcessing` → `validator({type:'sanity', ...})` → `sanity-zip` / `sanity-folder` → `false` → the API returns `401 VALIDATION_ERROR` for every upload. The old `'cmsType'` placeholder failed the same way, but it was self-evidently a placeholder; `'sanity'` reads as a supported CMS and will send people hunting in the wrong place.

This also makes the new archive branch in `routes/index.ts` unreachable-by-default in the success sense: it extracts to disk, then `sanity-folder` returns `false`, so it always answers 401 while leaving the extracted directory behind. Today only `aem` can consume a `folder` payload.

Suggested fix: revert this to the neutral placeholder (or drop the fallback and fail fast when `CMS_TYPE` is unset), and land the `sanity` validator/parser in the same PR if the archive path is meant to be usable.

---

upload-api/src/config/index.ts
```diff
@@ -58,5 +58,5 @@ export default {
   },
 
   // Local path for the CMS data
-  localPath: process.env.CMS_LOCAL_PATH || process.env.CONTAINER_PATH || 'your_local_cms_data_path',
+  localPath: process.env.CMS_LOCAL_PATH || process.env.CONTAINER_PATH || '/Users/umesh.more/Documents/sanity/backup-export.tar.gz',
```
umesh-more-cstk, line 61: **blocker:** A developer's personal absolute path is committed as the shipped default: `/Users/umesh.more/Documents/sanity/backup-export.tar.gz`.

Two concrete problems:

1. **Broken default everywhere else.** Any environment without `CMS_LOCAL_PATH` or `CONTAINER_PATH` (Docker, CI, another dev's laptop) now resolves `localPath` to a path that doesn't exist. `routes/index.ts` does `statSync(localPath)` in a try/catch and returns `500 Error accessing local path` — a 500 instead of a clear config error.
2. **It's exposed over the API.** `router.get('/config')` returns the whole config object to the client with only `mysql.password` stripped, and the file-processing responses embed `file_details: config` as well. So this leaks the author's OS username and home-directory layout to every UI client and into every error payload.

Suggested fix: restore the neutral placeholder (`'your_local_cms_data_path'`), or better, drop the literal fallback and let the route return an explicit "CMS_LOCAL_PATH is not configured" 400 when neither env var is set. Local paths belong in your `.env`, not in `config/index.ts`.

---

upload-api/src/helper/index.ts
```diff
@@ -129,6 +130,24 @@ const saveZip = async (zip: any, name: string) => {
...
+
+/**
+ * Extract a tar / tar.gz / tgz archive into `extracted_files/<destName>` and return that
+ * directory path. `tar.x` auto-detects gzip, so the same call handles compressed and plain tars.
+ */
+const extractArchive = async (archivePath: string, destName: string): Promise<string> => {
+  const baseDir = path.join(__dirname, '..', '..', 'extracted_files');
+  const destDir = path.join(baseDir, destName);
+  await fs.promises.mkdir(destDir, { recursive: true });
+  await tar.x({ file: archivePath, cwd: destDir });
```
umesh-more-cstk, line 147: **blocker:** `tar.x` runs with no entry filter and no size cap on an archive that is, by design, untrusted CMS export data.

To be precise about what is and isn't covered: node-tar 6 does defend against classic tar-slip on its own — with the default `preservePaths: false` it strips leading `/` from absolute entry paths and refuses entries containing `..`, and it won't write *through* an existing symlink. So `../../etc/foo` is not the gap here.

What is still unguarded:

1. **Symlink / hardlink entries get created inside `destDir`.** An export containing an entry `assets/creds -> /root/.aws/credentials` (or `-> /proc/self/environ`) extracts as a live symlink. The folder validator and `createMapper` then walk `extracted_files/<name>/` and read files through it — pulling host filesystem content into the migration payload that gets pushed to Contentstack. Reading *through* a symlink is not something node-tar protects against; that's on the consumer.
2. **Decompression bomb.** No cap on entry count or total extracted bytes. A few-hundred-KB `.tar.gz` can expand to hundreds of GB and fill the server disk; nothing here aborts partway.

Suggested fix:

```ts
await tar.x({
  file: archivePath,
  cwd: destDir,
  preservePaths: false,          // explicit, don't rely on the default
  filter: (_p, entry) => entry.type === 'File' || entry.type === 'Directory',
  onentry: (entry) => {
    total += entry.size;
    if (total > MAX_EXTRACT_BYTES) throw new Error('Archive exceeds maximum extracted size');
  }
});
```

with `total` reset per call and `MAX_EXTRACT_BYTES` in config. The `filter` is the important half — it drops `SymbolicLink` / `Link` entries outright.

---

upload-api/src/helper/index.ts
```diff
@@ -129,6 +130,24 @@ const saveZip = async (zip: any, name: string) => {
...
+};
+
+/**
+ * Extract a tar / tar.gz / tgz archive into `extracted_files/<destName>` and return that
+ * directory path. `tar.x` auto-detects gzip, so the same call handles compressed and plain tars.
+ */
+const extractArchive = async (archivePath: string, destName: string): Promise<string> => {
+  const baseDir = path.join(__dirname, '..', '..', 'extracted_files');
+  const destDir = path.join(baseDir, destName);
+  await fs.promises.mkdir(destDir, { recursive: true });
```
umesh-more-cstk, line 146: **blocker:** `mkdir({recursive:true})` + extract merges into a pre-existing directory instead of replacing it — stale data from a previous migration survives and gets migrated.

`destDir` is keyed only on the archive's base name, and tar overwrites the entries it carries but never removes files it doesn't. `createMapper`'s `clearAllMigrationData()` only sweeps folders matching `*MigrationData` at the upload-api root, so `extracted_files/<name>/` is never cleaned by anything.

Failure scenario: a user uploads `backup-export.tar.gz` containing `data/posts.ndjson` and `data/legacy.ndjson`, notices `legacy.ndjson` shouldn't be there, re-exports without it, and re-uploads the corrected `backup-export.tar.gz`. The second extraction leaves the old `legacy.ndjson` in place and the migration ingests deleted content as if it were live. Silent, and hard to trace back to extraction.

Suggested fix:

```ts
await fs.promises.rm(destDir, { recursive: true, force: true });
await fs.promises.mkdir(destDir, { recursive: true });
```

Worth also deleting `destDir` when validation fails downstream, so failed uploads don't accumulate on disk.

---

upload-api/src/routes/index.ts
```diff
@@ -184,6 +184,29 @@ router.get(
...
         const fileExt = fileName?.split('.')?.pop() ?? '';
 
+        // Archive exports (e.g. Sanity ships a .tar.gz) aren't a single parsable file:
+        // extract them to a folder and run the folder-based validator/mapper flow.
+        if (isArchive(fileName)) {
+          const safeName = sanitizeFilename(name);
+          const baseDir = path.join(__dirname, '..', '..', 'extracted_files');
+          const extractedDir = await extractArchive(localPath, safeName);
+
+          if (!isPathWithinBase(extractedDir, baseDir)) {
```
umesh-more-cstk, line 194: **blocker:** This path-traversal guard is dead code, and it runs after the extraction it's supposed to gate.

Trace it: `extractArchive` computes `destDir = path.join(baseDir, destName)` from the *same* `baseDir` this check uses, and `destName` is already `sanitizeFilename(name)`, which does `path.basename(...)` and strips everything outside `[a-zA-Z0-9_.\s-]` — so no separators, no `..`. `isPathWithinBase(extractedDir, baseDir)` is therefore structurally incapable of returning `false`. The 400 branch can never execute.

Two things follow:

1. **Ordering.** Even if the check could fire, line 192 has already written the archive contents to disk. A guard placed after the side effect it guards protects nothing.
2. **False assurance.** The comment says "Path traversal attempt detected", which reads like the archive's contents were validated. They weren't — the only thing checked is a path this function just built itself. See my note on `helper/index.ts:147` for the entry-level risk that is actually live.

Suggested fix: drop this block, validate `safeName` *before* calling `extractArchive` (see the `line 190` comment), and move the real defence into the `tar.x` options. Compare the XML branch at line 262 — there the check is meaningful because `filePath` incorporates externally-influenced input.

---

upload-api/src/routes/index.ts
```diff
@@ -184,6 +184,29 @@ router.get(
         const name = fileName?.split?.('.')?.[0];
         const fileExt = fileName?.split('.')?.pop() ?? '';
 
+        // Archive exports (e.g. Sanity ships a .tar.gz) aren't a single parsable file:
+        // extract them to a folder and run the folder-based validator/mapper flow.
+        if (isArchive(fileName)) {
+          const safeName = sanitizeFilename(name);
```
umesh-more-cstk, line 190: **blocker:** `safeName` can be the empty string, which makes the extraction target `extracted_files/` itself — the shared directory other flows write into.

`name` is `fileName.split('.')[0]` (line 185), and `sanitizeFilename` returns `''` for empty input and strips every character outside `[a-zA-Z0-9_.\s-]`. Both of these produce `safeName === ''`:

- `backup.tar.gz` renamed with a leading dot → basename `.tar.gz` → `split('.')[0]` is `''`
- a non-ASCII archive name, e.g. `导出.tar.gz` or `экспорт.tar.gz` → every character stripped → `''`

Then `path.join(baseDir, '')` collapses to `baseDir`, and the archive unpacks directly into `extracted_files/`. That's where `saveJson` writes `<name>.json` for the WordPress/Drupal flows — so a tar entry named `wordpress.json` silently overwrites another project's parsed export. The guard on line 194 won't catch it either: `isPathWithinBase(baseDir, baseDir)` yields a relative path of `''`, which passes.

Suggested fix, before the `extractArchive` call:

```ts
const safeName = sanitizeFilename(name);
if (!safeName) {
  return res.status(400).json({
    status: 400,
    message: 'Could not derive a safe name from the archive filename.',
    file_details: config
  });
}
```

A per-upload unique segment (e.g. the sanitized `projectId`) as the parent directory would also fix the cross-project collision more generally.

---

upload-api/src/routes/index.ts
```diff
@@ -184,6 +184,29 @@ router.get(
...
+          if (!isPathWithinBase(extractedDir, baseDir)) {
+            console.error('Path traversal attempt detected');
+            return res.status(400).json({
+              status: 400,
+              message: 'Invalid extraction path.',
+              file_details: config
+            });
+          }
+
+          const data = await handleFileProcessing('folder', extractedDir, cmsType, name);
```
umesh-more-cstk, line 203: **nit:** The unsanitized `name` is passed here while the directory was built from `safeName`, so the two can disagree (e.g. `my export!.tar.gz` → dir `my export.tar` vs. name `my export!`).

For the `folder` extension `handleFileProcessing` only forwards `data` to the validator and ignores `name`, so nothing breaks today — but if a Sanity folder validator ever uses the name to locate a manifest, it'll be looking for the wrong thing. Passing `safeName` costs nothing and keeps the two consistent.

---

umesh-more-cstk (commented):
## Review summary

**6 blockers, 1 nit.** Non-gating review — leaving the merge decision to a human.

The PR is 46 files: 40 are new `.claude/` skill + agent definitions (docs/prompt content, not reviewed as code), and 6 are real changes — a tar/tar.gz extraction path in `upload-api`, plus two config defaults. The findings are all in that second group.

### Blockers

| # | Location | Issue |
|---|---|---|
| 1 | `upload-api/src/config/index.ts:32` | Default `cmsType` set to `'sanity'`, but no Sanity validator exists — every upload 401s under defaults |
| 2 | `upload-api/src/config/index.ts:61` | Personal absolute path `/Users/umesh.more/...` committed as the default `localPath`, and it's served to clients by `GET /config` |
| 3 | `upload-api/src/helper/index.ts:147` | `tar.x` with no entry filter or size cap on untrusted archives — symlink entries and decompression bombs |
| 4 | `upload-api/src/helper/index.ts:146` | Extraction merges into an existing directory; stale files from a prior export survive and get migrated |
| 5 | `upload-api/src/routes/index.ts:194` | The path-traversal guard is unreachable and runs *after* extraction |
| 6 | `upload-api/src/routes/index.ts:190` | Empty `safeName` collapses the extraction target onto the shared `extracted_files/` root |

Blockers 1 and 5 compound: the archive branch extracts to disk, then `sanity-folder` fails validation, then nothing cleans up the extracted directory. As it stands the new code path cannot succeed end-to-end with the defaults this PR ships.

### Scope

The branch is `feat/skill` and the title is "Feat/skill", but the diff also flips two `upload-api` runtime config values and adds a tar-extraction route branch. Those look like leftovers from local Sanity testing rather than part of the skills work. Two options, both fine — split them into their own PR, or keep them and say so in the description. What doesn't work is shipping them unannounced.

Smaller scope notes:
- `add-cms-connector-workflow.svg` is at the repo root; it belongs under `.claude/skills/add-cms-connector/` with the material it documents.
- `upload-api/migration-aem/package-lock.json` (−7 lines) is unrelated churn.
- `upload-api/package-lock.json` is consistent with the `tar` addition (`tar`, `chownr`, `fs-minipass`, `minizlib`, `yallist`, plus `date-fns` re-hoisting). No content concerns.

### PR hygiene

- The description is the untouched template — placeholder `MIGRATION-XXXX` Jira link, no **PR Type** or **Affected Areas** boxes ticked, empty checklist. With a diff this size the Affected Areas checkboxes are the fastest way to catch exactly the scope mismatch above.
- `mergeable_state` is `dirty` — this conflicts with `dev` and needs a merge before it can land.
- The author checklist requires a `feature/` / `bugfix/` / `hotfix/` branch prefix; `feat/skill` doesn't match.
- No tests accompany the new extraction helper. `isArchive` and `extractArchive` are pure and easy to cover — worth a few cases (empty name, nested paths, symlink entry rejected once the filter lands).

### Not flagged

`createMapper(extractedDir, ...)` on line 205 is a floating promise, but the sibling folder and XML branches do the same thing — consistent with existing convention, so it's out of scope for this PR rather than a new defect.

---

PR #1140, opened 2026-08-07
fix(CMG-1100): migration failure no longer stalls the whole run

chetan-contentstack:
## 🔗 Jira Tickets

- [Delta Migration | v1.0.0 | Entry localization failed. Issue we are getting but on migration screen I see success message](https://contentstack.atlassian.net/browse/CMG-1100)

---

## 📋 PR Type

- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Hotfix
- [ ] ♻️ Refactor
- [ ] 📝 Documentation

---

### What changed?

- **`runCli.service.ts`** — the CLI import (`cm:stacks:import`) can hard-fail (bad auth, stack not found, etc.) without the code ever reaching the point where it writes the `Migration Process Completed` log line the UI waits for. The run would spin forever with no error and no way to retry. Now writes an explicit `Migration Process Failed` marker on any hard failure (non-zero exit or thrown exception) and resets the stuck `isMigrationStarted` flag so the run is retryable. Extracted a shared `writeFailureMarker()` helper used by both the reject path and the exception catch.
- **`entry-update-script.cjs`** — `updateEntryTask()` wrapped its entire per-content-type/per-entry update loop in a single try/catch that rethrew on any error. One entry failing (e.g. a Contentstack API validation error like "Entry localization failed") aborted every other entry queued behind it — confirmed live: a single bad `Date` value on one entry's secondary locale caused *zero* entries to get that locale, not just the bad one. Moved the try/catch inside the per-entry loop so a failure is logged and the loop continues, matching `updateAssetTask()`'s already-correct per-item handling right above it in the same file.
- **`entryAssetMapper.tsx`** — the Save-footer's containing box was sized with `height: calc(100vh - 246px)`, a hardcoded guess at the height of everything above it. Any drift (title length, zoom, browser chrome) silently pushed the sticky Save button below the viewport. Replaced with a `useMeasuredBoxHeight` hook that measures the box's real `getBoundingClientRect().top` on mount/resize and sets height to exactly `window.innerHeight - top`, so it self-corrects instead of relying on a constant.
- **`MigrationLogViewer.tsx`** — added detection for the new `Migration Process Failed` terminal message alongside the existing success check: shows an error toast and resets `migrationStarted` so Execute Migration becomes actionable again, instead of the UI spinning indefinitely.

### Why?

QA reported that an unsupported/invalid field value during migration leaves the run stuck with no error and no way to proceed. Root-caused to two independent silent-failure paths (CLI-level hard failure, and per-entry localization failure) plus one unrelated layout regression on the same screen found during manual testing.

---

## 🧩 Affected Areas

- [x] `api` — Node.js backend
- [x] `ui` — React frontend
- [ ] Other

---

## 🧪 How to Test

1. **Hard CLI failure:** trigger `Execute Migration` with an invalid/expired auth token or stack id so `cm:stacks:import` exits non-zero.
   - **Before:** UI spins forever, nothing in the log, no way to retry.
   - **After:** log shows `Migration Process Failed`, an error toast appears, and Execute Migration becomes clickable again.
2. **Per-entry localization failure:** set a `Date`-type field's secondary-locale value to a non-date string (e.g. `"not-a-valid-date"`) on one entry, migrate a content type with 2+ entries into 2+ locales.
   - **Before:** console shows `422 Entry localization failed`; *no* entry gets its secondary-locale version.
   - **After:** only the bad entry logs a failure (`Failed to update entry <uid> (locale "en-IN"): ...`); every other entry gets its secondary-locale version normally.
3. **Save button clipping:** open Map Entry (Entries or Assets tab) with only a couple of rows; confirm the Save button is fully visible and not cut off at the bottom of the viewport, including after a browser resize/zoom change.
4. `cd api && npm run test:coverage` — 742/742 passing, thresholds met.
5. `cd ui && npx vitest run` — 346/346 passing.

**Expected result:** genuine CLI failures surface an error and unlock retry; a single bad entry no longer blocks the rest of the batch; Save button stays visible on Map Entry regardless of viewport/zoom.

---

## 📸 Screenshots / Recordings

| Before | After |
|--------|-------|
|        |       |

---

- Found while manually testing [#1136](https://github.com/contentstack/migration-v2/pull/1136) (CMG-1101) — no code dependency, opened as a standalone fix against `dev`.

---

## ✅ Author Checklist

- [x] Branch follows naming convention: `bugfix/cmg-1100-migration-continue`
- [x] Self-reviewed the diff
- [x] No secrets committed
- [x] Existing tests pass (`npm run test:coverage` in `api/`, `vitest run` in `ui/`)
- [ ] New tests written — not added; `runCli.service.ts`'s CLI-spawn path, `entry-update-script.cjs`'s migration-runner entry point, and the layout hook aren't currently under this repo's test harness

---

## 👀 Reviewer Notes

- `runCli.service.ts` distinguishes two failure classes deliberately: a **hard failure** (non-zero exit / thrown exception — nothing happened) now marks the run failed and retryable; a **per-entry/per-task error inside an otherwise-successful exit** (the CLI's own best-effort behavior) is left as-is — still logged, but doesn't block completion. Don't conflate the two; an earlier iteration of this fix over-corrected and treated all errors as hard failures, which regressed "continues past a single bad entry."
- The `entry-update-script.cjs` fix intentionally mirrors `updateAssetTask()`'s existing pattern in the same file rather than introducing a new error-handling style.
- `useMeasuredBoxHeight` in `entryAssetMapper.tsx` measures `top`, not `height`, specifically to avoid a resize-observer feedback loop (observing and then resizing the same box's height would retrigger itself if it also measured height).

---

### DIFF
--- a/api/src/services/runCli.service.ts
+++ b/api/src/services/runCli.service.ts
@@ -89,7 +89,12 @@ const runCommand = (
       }
     });
 
-    // For stderr handler
+    // For stderr handler — a per-entry/per-task failure (e.g. an "Entry localization
+    // failed" 422 from the management API) surfaces here even when the CLI's overall
+    // import still exits 0 and continues with the rest of the batch. Log it as an error
+    // line so it's visible in the execution log, but don't treat it as fatal — the CLI is
+    // deliberately best-effort per entry, and the exit code is still the source of truth
+    // for whether the run as a whole succeeded or needs a retry.
     cmdProcess.stderr.on('data', (data) => {
       const output = data.toString();
       process.stderr.write(output); // Keep colors in console
@@ -130,6 +135,47 @@ const runCommand = (
       }
     });
   });
+};
+
+/**
+ * Writes a distinct failure marker to the migration log(s) and clears the stuck
+ * "in progress" flag, so MigrationLogViewer.tsx — which only ever stops waiting on an
+ * exact terminal string — doesn't spin forever, and the run can be retried.
+ */
+const writeFailureMarker = async (
+  isTest: boolean,
+  projectId: string,
+  transformePath: string,
+  loggerPath?: string
+): Promise<void> => {
+  try {
+    const failureLogEntry = {
+      level: 'error',
+      message: isTest ? 'Test Migration Process Failed' : 'Migration Process Failed',
+      timestamp: new Date().toISOString(),
+    };
+    fs.appendFileSync(transformePath, JSON.stringify(failureLogEntry) + '\n');
+    if (loggerPath && loggerPath !== transformePath) {
+      fs.appendFileSync(loggerPath, JSON.stringify(failureLogEntry) + '\n');
+    }
+  } catch (logErr) {
+    console.error('Error writing failure marker to log file:', logErr);
+  }
+  if (!isTest) {
+    try {
+      await ProjectModelLowdb.read();
+      const projectIndex = ProjectModelLowdb.chain
+        .get('projects')
+        .findIndex({ id: projectId })
+        .value();
+      if (projectIndex > -1) {
+        ProjectModelLowdb.data.projects[projectIndex].isMigrationStarted = false;
+        await ProjectModelLowdb.write();
+      }
+    } catch (statusErr) {
+      console.error('Error resetting migration status after failure:', statusErr);
+    }
+  }
 };
 
 /**
@@ -234,7 +280,16 @@ export const runCli = async (
         transformePath
       ); // Pass the log file path here
 
-      // After the import command completes
+      // After the import command completes.
+      //
+      // Note: the CLI's import plugin runs best-effort per entry — a single entry failing
+      // (e.g. an "Entry localization failed" 422, logged as an 'error' line above by the
+      // stderr handler) does not stop it from continuing with and completing the rest of the
+      // batch, and it still exits 0. That's by design, so a run is treated as complete here
+      // whenever the process exits 0, even if some individual entries logged errors along the
+      // way — those errors remain visible in the execution log for follow-up, but don't block
+      // the rest of the migration. Only a non-zero exit (handled below, in the catch block)
+      // means nothing happened and the run needs a full retry.
 
       // Write the completion message ONCE in the format the UI expects
       if (isTest) {
@@ -280,11 +335,6 @@ export const runCli = async (
         await writeUidMapping(backupPath, projectId, iteration);
         await writePerLocaleEntryUidMapping(backupPath, projectId, iteration);
       }
-
-      // Keep the project status update code:
-      // ... rest of the code ...
-
-      // Add debug logs to track project index and test flag
 
       // Make sure we have the latest data
       await ProjectModelLowdb.read();
@@ -350,6 +400,12 @@ export const runCli = async (
     }
   } catch (error) {
     console.error('🚀 ~ runCli ~ error:', error);
+    // The CLI import can hard-fail (e.g. a stack import validation error) after
+    // `runCommand` already logged the raw error, but nothing below this point ever ran —
+    // in particular the 'Migration Process Completed' terminal marker never got written.
+    // MigrationLogViewer.tsx only ever stops waiting when it sees that exact string, so
+    // without an explicit failure marker the UI spins forever with no way to retry.
+    await writeFailureMarker(isTest, projectId, transformePath);
   }
 };
 
--- a/api/src/services/runCli.service.ts
+++ b/api/src/services/runCli.service.ts
@@ -145,8 +145,7 @@ const runCommand = (
 const writeFailureMarker = async (
   isTest: boolean,
   projectId: string,
-  transformePath: string,
-  loggerPath?: string
+  transformePath: string
 ): Promise<void> => {
   try {
     const failureLogEntry = {
@@ -155,9 +154,6 @@ const writeFailureMarker = async (
       timestamp: new Date().toISOString(),
     };
     fs.appendFileSync(transformePath, JSON.stringify(failureLogEntry) + '\n');
-    if (loggerPath && loggerPath !== transformePath) {
-      fs.appendFileSync(loggerPath, JSON.stringify(failureLogEntry) + '\n');
-    }
   } catch (logErr) {
     console.error('Error writing failure marker to log file:', logErr);
   }
@@ -291,52 +287,15 @@ export const runCli = async (
       // the rest of the migration. Only a non-zero exit (handled below, in the catch block)
       // means nothing happened and the run needs a full retry.
 
-      // Write the completion message ONCE in the format the UI expects
-      if (isTest) {
-        const directLogEntry = {
-          level: 'info',
-          message: 'Test Migration Process Completed',
-          timestamp: new Date().toISOString(),
-        };
-
-        // Write to the transform path (main log file) - ONLY ONCE
-        fs.appendFileSync(
-          transformePath,
-          JSON.stringify(directLogEntry) + '\n'
-        );
-
-        // Also write to backup log path if different
-        if (loggerPath && loggerPath !== transformePath) {
-          fs.appendFileSync(loggerPath, JSON.stringify(directLogEntry) + '\n');
-        }
-      } else {
-        const directLogEntry = {
-          level: 'info',
-          message: 'Migration Process Completed',
-          timestamp: new Date().toISOString(),
-        };
-
-        // Write to the transform path (main log file) - ONLY ONCE
-        fs.appendFileSync(
-          transformePath,
-          JSON.stringify(directLogEntry) + '\n'
-        );
-
-        // Also write to backup log path if different
-        if (loggerPath && loggerPath !== transformePath) {
-          fs.appendFileSync(loggerPath, JSON.stringify(directLogEntry) + '\n');
-        }
-        await ProjectModelLowdb.read();
-        const projectData = ProjectModelLowdb.chain
-          .get("projects")
-          .find({ id: projectId })
-          .value();
-        const iteration = projectData?.iteration || 1;
-        await writeUidMapping(backupPath, projectId, iteration);
-        await writePerLocaleEntryUidMapping(backupPath, projectId, iteration);
-      }
-
-      // Make sure we have the latest data
+      // Do all post-import bookkeeping BEFORE writing the completion marker. If any of it
+      // throws, execution falls straight to the catch block below, which is then the only
+      // thing that ever writes a terminal marker for this run — so the log can never end up
+      // with both 'Completed' and 'Failed' for the same run. (It used to: the marker was
+      // written first, so a later throw here — e.g. from writeUidMapping or a lowdb write —
+      // would append 'Failed' after 'Completed' was already there, and
+      // MigrationLogViewer.tsx fires both its success and failure effects in the same pass,
+      // leaving migrationCompleted's final value dependent on dispatch order even though the
+      // backend had already persisted `isMigrationCompleted: true`.)
       await ProjectModelLowdb.read();
       const projectIndex = ProjectModelLowdb.chain
         .get('projects')
@@ -364,6 +323,10 @@ export const runCli = async (
 
       // Update project status for non-test migrations
       if (projectIndex > -1 && !isTest) {
+        const iteration = ProjectModelLowdb.data.projects[projectIndex]?.iteration || 1;
+        await writeUidMapping(backupPath, projectId, iteration);
+        await writePerLocaleEntryUidMapping(backupPath, projectId, iteration);
+
         // Direct modification might be more reliable
         ProjectModelLowdb.data.projects[projectIndex].isMigrationCompleted =
           true;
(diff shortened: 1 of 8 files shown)


api/src/services/runCli.service.ts
```diff
@@ -132,6 +137,47 @@ const runCommand = (
...
+const writeFailureMarker = async (
+  isTest: boolean,
+  projectId: string,
+  transformePath: string,
+  loggerPath?: string
+): Promise<void> => {
+  try {
+    const failureLogEntry = {
+      level: 'error',
+      message: isTest ? 'Test Migration Process Failed' : 'Migration Process Failed',
```
umesh-more-cstk, line 153: **blocker:** `Test Migration Process Failed` is written but nothing ever reads it, so the test-migration path still hangs exactly as before.

Test-migration logs are consumed by `ui/src/components/LogScreen/index.tsx`, not by `MigrationLogViewer.tsx`. That component's only terminal check is `index.tsx:177` (`message === 'Test Migration Process Completed'`), which is what clears `setisLogsLoading(false)` and writes `isTestMigrationStarted: false` to local storage. A grep for the failure string across the repo returns exactly one hit — this line. There is no consumer.

Failure scenario: user runs a **test** migration against a stack with bad auth / a deleted stack. `runCli` throws, this marker is appended, and `LogScreen/index.tsx` never matches it → `isLogsLoading` stays `true`, the spinner runs forever, and `testmigration_<projectId>` keeps `isTestMigrationStarted: true` in local storage, so the stuck state survives a page reload. That's the same bug this PR is fixing for the real path, still live on the test path.

Note the second half too: `writeFailureMarker` deliberately skips the status reset when `isTest`, but the test path's "in progress" flag lives in local storage rather than lowdb — so nothing clears it either.

Suggested fix: add a failure branch to `LogScreen/index.tsx` alongside the completion branch at :177 that clears `isLogsLoading`, writes `{ isTestMigrationCompleted: false, isTestMigrationStarted: false }` to local storage, calls `sendDataToParent?.(false)`, and shows an error notification.

---

api/src/services/runCli.service.ts
```diff
@@ -350,6 +400,12 @@ export const runCli = async (
     }
   } catch (error) {
     console.error('🚀 ~ runCli ~ error:', error);
+    // The CLI import can hard-fail (e.g. a stack import validation error) after
+    // `runCommand` already logged the raw error, but nothing below this point ever ran —
+    // in particular the 'Migration Process Completed' terminal marker never got written.
+    // MigrationLogViewer.tsx only ever stops waiting when it sees that exact string, so
+    // without an explicit failure marker the UI spins forever with no way to retry.
+    await writeFailureMarker(isTest, projectId, transformePath);
```
umesh-more-cstk, line 383: **blocker:** this catch can't tell "failed before completion" from "failed after completion", so a post-completion error writes *both* markers and lands the UI in contradictory state.

The completion marker is written at :315-328, but the `try` keeps running well past it — `writeUidMapping` / `writePerLocaleEntryUidMapping` (:335-336), the lowdb status writes (:340-376), and `recordMigratedLocales` (:395). Anything throwing in that tail reaches this catch, which unconditionally appends `Migration Process Failed` to a log that already contains `Migration Process Completed`.

Failure scenario: iteration-1 non-test run, import exits 0, completion marker written, then `writeUidMapping` throws (bad backup dir / EACCES / malformed mapper file). `MigrationLogViewer.tsx` scans the whole `logs` array for both strings in the *same* effect pass, so both branches fire: the completion branch (:221) opens the success modal and dispatches `migrationCompleted: true`, then the failure branch (:268) dispatches `migrationCompleted: false` and fires an error toast. Last dispatch wins → user sees a success modal *and* an error toast, with `migrationCompleted` false. Backend disagrees too: `isMigrationCompleted: true` was already persisted at :368 before the throw.

Suggested fix: track whether the completion marker was written and make the failure marker conditional, e.g. set `let completionMarkerWritten = false` before the `if (isTest)` block, flip it to `true` right after the `appendFileSync` calls, and here do `if (!completionMarkerWritten) await writeFailureMarker(...)`. Alternatively wrap the post-completion bookkeeping (:335-396) in its own try/catch so those errors log without being mistaken for a failed migration.

---

api/src/utils/entry-update-script.cjs
```diff
@@ -333,13 +337,16 @@ module.exports = async ({
...
+                            // updateAssetTask's per-item error handling above.
+                            console.error(`Failed to update entry ${realEntryUid}${locale ? ` (locale "${locale}")` : ''}:`, error?.message || error);
                         }
                     }
-                    console.info('All entries updated successfully');
-                } catch (error) {
-                    console.error(error);
-                    throw error;
                 }
+                console.info('All entries processed');
```
umesh-more-cstk: **blocker:** moving the try/catch per-entry is the right fix, but the task now *never* throws — so a delta run where every entry failed reports success to the user.

The per-entry containment itself is correct and does match `updateAssetTask`. The problem is the signal that got dropped with the old rethrow. Tracing the path:

- The task no longer throws → `cm:stacks:migration` exits 0 → `runCommand` in `updateEntryCli.service.ts:202` resolves.
- So execution reaches :233-239, which appends `Entry Update Process Completed`, instead of falling into the catch at :240 that writes `Failed to update entries for stack`.
- `MigrationLogViewer.tsx:214-221` treats `Entry Update Process Completed` as *the* terminal success message on iteration ≥ 2 → success modal, green notification.
- `migration.service.ts:1292` then calls `recordDeltaMigratedLocales` regardless, and that derives `ranLocales` from the update-config / uid-mapper contents (:1388-1394) — not from whether any update actually landed.

Failure scenario: delta run where the management API rejects every entry (expired token, locale not on the stack, a validation error affecting a shared field). Every entry hits the new catch and is logged; the task completes; the UI shows the success modal; the locales get recorded as migrated, so the next delta iteration skips those entries entirely. Silent data loss, and the user has no signal beyond stderr lines buried in the log. This is the same class of bug the comment at `runCli.service.ts:381-385` warns about.

Also note `failedMessage: "Failed to update entries"` is now unreachable.

Suggested fix: count failures and throw once *after* the loops, which keeps the "don't abort siblings" fix while restoring the signal:

```js
let failed = 0;
// ...in the catch: failed++;
// after both for-loops:
console.info(`Entry update finished — ${failed} failed`);
if (failed > 0) throw new Error(`Failed to update ${failed} entr${failed === 1 ? 'y' : 'ies'}`);
```

If a partial failure genuinely shouldn't fail the run, then it should at least be surfaced explicitly rather than reported as success — and `recordDeltaMigratedLocales` should exclude locales whose entries failed.

---

ui/src/components/ContentMapper/entryAssetMapper.tsx
```diff
@@ -13,6 +13,35 @@ interface entryAssetMapperProps {
...
+
+  useEffect(() => {
+    const measure = () => {
+      const top = ref.current?.getBoundingClientRect().top;
+      if (top == null) return;
+      setHeight(Math.max(0, window.innerHeight - top));
+    };
+
+    measure();
+    const ro = new ResizeObserver(measure);
```
umesh-more-cstk, line 37: **question:** the ResizeObserver watches the box itself, which is the one element whose size this hook already controls — so it won't catch the drift the comment says it's for.

The stated trigger is "the stepper/title chrome changes height". But once `measure()` has run, the box carries an inline fixed `height`/`maxHeight`. If chrome *above* it grows, the box's top offset changes while its own border-box size does not — no RO callback fires, and the height stays stale. Mount-time measurement and `window.resize` (which covers zoom) do work; it's the post-mount chrome change that slips through.

Failure scenario: a validation banner or a wrapped long project title appears above this box after mount without a window resize — the box keeps its old height and the sticky Save footer is pushed below the fold again, which is the original bug.

Suggested fix: observe what actually moves the box rather than the box itself, e.g. `ro.observe(document.body)` (or the parent/chrome element), keeping the `window.resize` listener as-is.

Minor secondary point: observing the element whose height you set inside the callback is a self-triggering loop. It converges here because the second pass computes the same value and React bails out of the re-render, but it can still surface "ResizeObserver loop completed with undelivered notifications" in the console. Switching the observed target resolves that too.

---

api/src/services/runCli.service.ts
```diff
@@ -132,6 +137,47 @@ const runCommand = (
...
+/**
+ * Writes a distinct failure marker to the migration log(s) and clears the stuck
+ * "in progress" flag, so MigrationLogViewer.tsx — which only ever stops waiting on an
+ * exact terminal string — doesn't spin forever, and the run can be retried.
+ */
+const writeFailureMarker = async (
+  isTest: boolean,
+  projectId: string,
+  transformePath: string,
+  loggerPath?: string
```
umesh-more-cstk: **nit:** `loggerPath` is never passed by the only caller, so this parameter and the `appendFileSync` at :159 are dead code.

`loggerPath` is declared at :250, inside the `try` block, so it is genuinely out of scope at the `catch` on :401 — the call at :408 can only pass three args. The effect is that the failure marker lands in `transformePath` only, while both completion markers (:309-311, :326-328) are mirrored into `loggerPath` as well. The socket stream the UI reads is fed from the path passed in as `serverPath`, so this doesn't break the fix — the backup log is just missing the failure line that its completion counterpart has.

Suggested fix: hoist `const loggerPath` above the `try` so the catch can forward it, or drop the parameter and the :158-160 branch to match what's actually reachable.

---

umesh-more-cstk (commented):
## Review — CMG-1100

Both underlying diagnoses are correct and well-evidenced: the CLI really can hard-fail without writing a terminal marker, and `updateEntryTask`'s single outer try/catch really did abort every entry queued behind the first failure. The per-entry containment genuinely matches `updateAssetTask`. The comments explaining *why* each change is needed are unusually good and made this reviewable.

Three blockers, all about failure **signalling** rather than the fixes themselves — in each case a failure is now contained but not surfaced, which in two of them means the UI reports success:

1. **`runCli.service.ts:154`** — the test-migration marker (`Test Migration Process Failed`) has no consumer anywhere. Test migrations are read by `LogScreen/index.tsx`, which only matches the *Completed* string, so a failing test migration still spins forever and keeps `isTestMigrationStarted: true` in local storage across reloads. The PR's core fix doesn't reach the test path.
2. **`runCli.service.ts:408`** — the catch covers the tail of the `try` that runs *after* the completion marker is written (`writeUidMapping`, lowdb writes, `recordMigratedLocales`). A throw there appends `Migration Process Failed` to a log that already has `Migration Process Completed`, and `MigrationLogViewer` fires both branches in one effect pass → success modal plus error toast, with `migrationCompleted` ending up `false` while the backend has already persisted `true`.
3. **`entry-update-script.cjs:349`** — with the rethrow gone the task can no longer fail, so the CLI exits 0, `Entry Update Process Completed` gets written, and a delta run in which *every* entry failed shows the success modal. `recordDeltaMigratedLocales` then marks those locales migrated from the config contents, so the next delta iteration skips them. Suggested a post-loop failure count + single throw, which keeps the sibling-containment fix intact.

Plus one question on the `ResizeObserver` in `entryAssetMapper.tsx:32` (it observes the box whose height the hook itself sets, so the chrome-height drift it targets won't retrigger it), and one nit on the dead `loggerPath` parameter.

**Scope:** matches the description. The `MigrationLogViewer.tsx` change isn't called out in the summary but is the direct UI counterpart to the new marker, so it belongs. The `entryAssetMapper` layout fix is unrelated to CMG-1100 but is declared in the description — flagging only so it's a conscious call.

Not verified: I read the code but didn't run the suites, so the 742/742 and 346/346 claims are taken at face value. Worth confirming whether existing tests assert on `updateEntryTask` rethrowing — finding 3's suggested fix would change that contract.

---

ui/src/components/ContentMapper/entryAssetMapper.tsx
umesh-more-cstk: Following up on the Reviewer Note in the description:

> `useMeasuredBoxHeight` measures `top`, not `height`, specifically to avoid a resize-observer feedback loop

That's correct, and it does resolve the *secondary* half of this comment — measuring `top` is why the loop converges instead of oscillating. I want to flag that it doesn't resolve the primary half, so this doesn't get closed out on that basis.

The two are separate concerns:

- **Feedback loop** — solved by measuring `top` instead of `height`. Agreed, no issue.
- **Wrong element observed** — not affected by what you measure. `ro.observe(ref.current)` fires on changes to *this box's* border-box size. The scenario the hook targets is chrome *above* the box changing height, which moves the box's `top` while leaving its own size untouched (the more so now that the hook pins an inline `height`/`maxHeight` on it). No size change on the observed element means no callback, so the stale height persists.

So the covered cases are mount and `window.resize` (which does include zoom). The gap is a post-mount chrome height change with no window resize — a banner appearing, or a title reflowing to two lines.

`ro.observe(document.body)` instead of the box keeps the `top`-measuring approach exactly as-is and closes that gap. Happy to be wrong if there's a reason the chrome above can't change height after mount on this screen.

---

vikrantraut-cstk (approved):

